PR #93, opened 2022-09-28
chore: inline k3d script values

a-cordier:
The goal is to be all setup by just running 

```shell
curl -s https://raw.githubusercontent.com/gravitee-io/gravitee-kubernetes-operator/chore-inline-k3d-script/scripts/k3d.sh| bash
```


ThibaudAV (approved):
I still think it's not easy readable to have everything like this and difficult to maintain 🤷‍♂️ 

you didn't want to add the wait ready at the end ?

PR #128, opened 2022-10-28
chore: move k3d script to zx

a-cordier:


a-cordier:
Note: I'm not sure about the last commit, it does not look to make us earn that much in terms of timing

a-cordier:
Note: documentation update needed (the curl would need to be piped to zx instead of bash)

ThibaudAV (approved):
Shouldn't we keep the sh script that runs the zx? with a npx or something? 
there's also the doc to change, no?

PR #136, opened 2022-11-03
feat: admission webhooks

a-cordier:


a-cordier:
waiting for progress on https://github.com/kubernetes-sigs/controller-runtime/pull/2014 to return warnings as required

a-cordier:
Note: activating web hooks requires installing cert-manager on the cluster

see https://cert-manager.io/docs/installation/

PR #210, opened 2022-12-28
feat: allow multiple api contexts with templating

a-cordier:
see https://gravitee.atlassian.net/browse/APIM-290

See also https://github.com/gravitee-io/gravitee-api-management/pull/2874 for APIM compatibility with the new API model

Note: I think we should build a compatibility matrix with APIM starting from now


test/internal/fixtures.go
```diff
@@ -117,9 +119,9 @@ func (f *FixtureGenerator) NewApiDefinition(
 }
 
 func (f *FixtureGenerator) NewManagementContext(
```
kamiiiel, line 121: Method hasn't been renamed?

test/apidefinition_controller_error_test.go
```diff
@@ -28,20 +28,20 @@ limitations under the License.
...
 	gio "github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
-	"github.com/gravitee-io/gravitee-kubernetes-operator/pkg/keys"
 	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal"
 )
 
 var _ = Describe("Checking NoneRecoverable && Recoverable error", Label("DisableSmokeExpect"), func() {
 
 	Context("With basic ApiDefinition & ManagementContext", func() {
-		var managementContextFixture *gio.ManagementContext
+		var managementContextFixture *gio.ApiContext
```
kamiiiel, line 44: shall we keep the old variable name?

test/internal/fixtures.go
a-cordier: good catch, i've added a commit with all changes related to naming of the API context in tests

controllers/apim/apidefinition/internal/delegate.go
```diff
@@ -16,50 +16,135 @@ package internal
...
+		context, err := d.resolveContext(ref)
+
+		if err != nil {
+			d.log.Error(err, "Unable to resolve context ", "namespace", ref.Namespace, "name", ref.Name)
+			continue
+		}
+
+		d.addContext(context)
+
+		if api.IsMissingDeletionFinalizer() {
```
kamiiiel, line 64: Is here a reason to put this inside this for loop? This is not suppose to be for the ApiDefinition itself?

api/v1alpha1/apidefinition_types.go
```diff
@@ -33,33 +34,39 @@ type ApiDefinitionSpec struct {
 
 	// The contextRef refers to the namespace and the name of a ManagementContext used for
```
kamiiiel, line 35: name of an "ApiContext"...

controllers/apim/apidefinition/internal/delegate.go
a-cordier: 👍

api/model/namespaced_name.go
```diff
@@ -21,6 +21,14 @@ type NamespacedName struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+func NewNamespacedName(namespace, name string) NamespacedName {
+	return NamespacedName{Namespace: namespace, Name: name}
+}
+
 func (n NamespacedName) ToK8sType() types.NamespacedName {
```
kamiiiel, line 28: pass the reference seems better in here, same as line 32

a-cordier: I tried to use a pointer receiver but it tends to make the object harder to use

eg
```golang
func (api *ApiDefinition) GetOrGenerateCrossID() string {
	if api.Spec.CrossID != "" {
		return api.Spec.CrossID
	}

	return uuid.FromStrings(api.GetNamespacedName().String()) // <= This is not possible with a pointer receiver
}
```

a-cordier: Resolved as discussed

kamiiiel (approved):

internal/apim/service/applications.go
```diff
@@ -0,0 +1,41 @@
...
+type Applications struct {
+	*client.Client
+}
+
+func NewApplications(client *client.Client) *Applications {
+	return &Applications{Client: client}
+}
+
+func (svc *Applications) Search(query string, status string) ([]model.Application, error) {
+	url := svc.EnvTarget("/applications").WithQueryParam("query", query).WithQueryParam("status", status)
```
Okhelifi, line 33:
```suggestion
	url := svc.EnvTarget("applications").WithQueryParam("query", query).WithQueryParam("status", status)
```

Okhelifi (approved):

PR #213, opened 2023-01-13
ci: helm chart release

a-cordier:


.circleci/config.yml
```diff
@@ -356,6 +417,38 @@ workflows:
           requires:
             - Lint Helm chart
 
+  deploy:
+    when:
+      and:
+        - or:
+            - equal: [ci-staging, << pipeline.git.branch >>]
```
Okhelifi, line 420: To remove before merging on alpha ?

.circleci/config.yml
```diff
@@ -270,65 +387,85 @@ workflows:
...
-          filters:
-            branches:
-              only:
-                - master
-      - login-to-azure:
-          name: K8s login on apim-preprod and get ~/.kube/config
+            - Test operator
+            - Test chart
+            - Authenticate on cluster
+      - helm-install:
```
Okhelifi, line 440: Should we add an approval button on this one ? or do we consider that all building pull request will deploy something and if I want to have a specific version I deploy it manually using a command line on the cluster

Okhelifi (approved):
Good job 👍

config/rbac/role.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
```
kamiiiel, line 3: I was wondering if we only want to listen to a single namespace, do we still need ClusterRole and ClusterRoleBinding?
Same thing applies for other files in this folder

.circleci/config.yml
```diff
@@ -207,16 +259,16 @@ jobs:
...
+          var-name: DOCKERHUB_BOT_USER_TOKEN
       - run:
           name: Docker login
           command: docker login --username="${DOCKERHUB_BOT_USER_NAME}" -p="${DOCKERHUB_BOT_USER_TOKEN}"
       - run:
-          name: Install operator SDK 
+          name: Install operator SDK
           command: |
             export OPERATOR_SDK_RELEASE_VERSION=v1.23.0
             export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
```
kamiiiel, line 274: Maybe amd64|x86_64) echo -n amd64

config/rbac/role.yaml
a-cordier: Yes, @sebdevaux had the same thought

.circleci/config.yml
Okhelifi: Done

PR #237, opened 2023-02-08
Publish apis using ingress [ Part 1 ]

Okhelifi:
## Description

This pull request contains the creation of an ApiDefinition using an Ingress without referencing a template in the annotations.
Here we can create update and delete the Ingress. If the Ingress is successfully updated the ApiDefinition is updated to according to the modifications on the Path for example.

There is still some remaining questions around the Host and the gateway thats why the Api won't be callable through the APIM Gateway for the moment


controllers/apim/ingress/ingress_controller.go
```diff
@@ -0,0 +1,121 @@
...
+// Reconciler watches and reconciles Ingress objects.
+type Reconciler struct {
+	client.Client
+	Log      logr.Logger
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
+}
+
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses,verbs=get;list;watch
```
a-cordier, line 44: wondering if the rbac annotation to allow creating apidefinitions should not be defined here. But I guess we will be able to see this when merged

a-cordier:
wondering if [this file](https://github.com/gravitee-io/gravitee-kubernetes-operator/blob/alpha/controllers/apim/apidefinition/internal/import_api_definition_template.go) should be remove as part of the P.R. 🤔

test/ingress_controller_delete_test.go
```diff
@@ -0,0 +1,91 @@
...
+			api := &gio.ApiDefinition{}
+			Eventually(func() error {
+				err := k8sClient.Get(ctx, ingressLookupKey, api)
+				return err
+			}, timeout, interval).ShouldNot(Succeed())
+		})
+
+		AfterEach(func() {
+			By("Deleting the Ingress")
+			Expect(k8sClient.DeleteAllOf(
```
a-cordier, line 83: do we still need this as we delete after the suite runs ?

test/suite_test.go
```diff
@@ -137,6 +150,12 @@ var _ = SynchronizedBeforeSuite(func() {
 var _ = SynchronizedAfterSuite(func() {
 	By("Tearing down the test environment")
 
+	Expect(k8sClient.DeleteAllOf(
```
a-cordier, line 153: Do we still need this if we delete after the suite runs ?

I guess we could save a few ⏲️ by removing it

test/ingress_controller_create_test.go
```diff
@@ -0,0 +1,129 @@
...
+						Name:   "foo-svc",
+						Target: "http://foo-svc.default.svc.cluster.local:8080",
+					},
+				},
+			))
+		})
+
+		AfterEach(func() {
+			By("Deleting the Ingress")
+			Expect(k8sClient.DeleteAllOf(
```
a-cordier, line 121: Do we still need this if we delete after the suite runs ?

I guess we could save a few ⏲️ by removing it

controllers/apim/ingress/internal/update.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/gravitee-io/gravitee-kubernetes-operator/pkg/keys"
+	v1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+	util "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (d *Delegate) AddFinalizer(desired *v1.Ingress) error {
+	ingress := &v1.Ingress{}
+	if err := d.k8s.Get(
```
a-cordier, line 26: Is the get absolutely necessary here ? (Fair question)

controllers/apim/ingress/internal/api_definition_template.go
```diff
@@ -0,0 +1,112 @@
...
+			return nil, err
+		}
+	} else {
+		apiDefinition = DefaultApiDefinitionTemplate()
+	}
+
+	return MergeApiDefinition(d.log, apiDefinition, ingress), nil
+}
+
+func DefaultApiDefinitionTemplate() *v1alpha1.ApiDefinition {
```
a-cordier, line 47: May be we should leave the method un-exported ?

controllers/apim/ingress/internal/api_definition_template.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/go-logr/logr"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (d *Delegate) ResolveApiDefinitionTemplate(ingress *v1.Ingress) (*v1alpha1.ApiDefinition, error) {
+	var apiDefinition *v1alpha1.ApiDefinition
+	name, hasIngressTemplateAnnotation := ingress.Annotations[keys.IngressTemplateAnnotation]
```
a-cordier, line 32: there's a go pattern that uses ok has a variable name on map access, wdyt ?

see [comma ok idiom](https://www.golinuxcloud.com/go-comma-ok-idiom/)

controllers/apim/ingress/internal/api_definition_template.go
```diff
@@ -0,0 +1,112 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+// MergeApiDefinition
+// Transform the ingress as an API Definition as per https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource
+func MergeApiDefinition(
```
a-cordier, line 66: May be we should leave the method un-exported ?

controllers/apim/ingress/internal/api_definition.go
```diff
@@ -0,0 +1,97 @@
...
+	existingApiDefinition, err = d.GetApiDefinition(types.NamespacedName{Namespace: ingress.Namespace, Name: ingress.Name})
+	if err != nil {
+		return util.OperationResultNone, err
+	}
+
+	apiDefinition.Spec.DeepCopyInto(&existingApiDefinition.Spec)
+	return util.OperationResultUpdated, d.k8s.Update(d.ctx, existingApiDefinition)
+}
+
+func (d *Delegate) GetApiDefinition(key client.ObjectKey) (*v1alpha1.ApiDefinition, error) {
```
a-cordier, line 90: May be we should leave the method un-exported ?

controllers/apim/ingress/internal/api_definition.go
```diff
@@ -0,0 +1,97 @@
...
+
+	d.log.Info(
+		"No change detected on ApiDefinition. Skipped.",
+		"name", apiDefinition.Name,
+		"namespace", apiDefinition.Namespace,
+	)
+	return util.OperationResultNone, nil
+}
+
+func (d *Delegate) CreateApiDefinition(
```
a-cordier, line 58: May be we should leave the method un-exported ?

controllers/apim/ingress/ingress_controller.go
```diff
@@ -0,0 +1,121 @@
...
+	log := log.FromContext(ctx)
+
+	ingress := &netV1.Ingress{}
+	if err := r.Get(ctx, req.NamespacedName, ingress); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	d := internal.NewDelegate(ctx, r.Client, log)
+
+	if !ingress.DeletionTimestamp.IsZero() {
```
a-cordier, line 58: how about having something similar to the api controller ?

```golang
if !ingress.DeletionTimestamp.IsZero()() {
  reconcileErr = events.Record(event.Delete, ingress, func() error {
	  return delegate.Delete(ingress)
  })
  } else {
  reconcileErr = events.Record(event.Update, ingress, func() error {
	  return delegate.CreateOrUpdate(ingress)
  })
}
```

a-cordier: as a matter of a fact, this even tends to make me want to remove the `IsBeingDeleted` method of API definition for even more consistency

a-cordier:
👍 aside from the little non-blocking comments here and there

controllers/apim/ingress/internal/update.go
Okhelifi: Here it's to avoid the errors saying that we are trying to update an object that has been modified by the ApiDefinition controller.

controllers/apim/ingress/ingress_controller.go
Okhelifi: Yes after last refactor we did together, we removed it and I forget to push it again but you're right I'll do it

controllers/apim/ingress/internal/api_definition_template.go
Okhelifi: Done

a-cordier (approved):

PR #242, opened 2023-02-10
ci: enhance helm release

a-cordier:
The goal of the P.R. is to

  - generate the legacy `bundle.yml` file on releases in order to allow users that have already installed this way to upgrade without breaking anything
  - set chart version according to the release version
  - generate chart reference on release

### DIFF
--- a/helm/gko/tests/bundle_test.yaml
+++ b/helm/gko/tests/bundle_test.yaml
@@ -21,7 +21,7 @@ tests:
       - hasDocuments:
           count: 12
 
-  - it: Should have manager the release namespace resource
+  - it: Should have manager namespace resource
     documentIndex: 0
     asserts:
       - isKind:
--- a/helm/gko/tests/bundle_test.yaml
+++ b/helm/gko/tests/bundle_test.yaml
@@ -19,10 +19,21 @@ tests:
   - it: Should nest all needed resources
     asserts:
       - hasDocuments:
-          count: 11
+          count: 12
+
+  - it: Should have manager the release namespace resource
+    documentIndex: 0
+    asserts:
+      - isKind:
+          of: Namespace
+      - isAPIVersion:
+          of: v1
+      - equal:
+          path: metadata.name
+          value: NAMESPACE
 
   - it: Should have manager service account in release namespace
-    documentIndex: 0
+    documentIndex: 1
     asserts:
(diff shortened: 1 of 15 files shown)


helm/gko/tests/bundle_test.yaml
```diff
@@ -19,10 +19,21 @@ tests:
   - it: Should nest all needed resources
     asserts:
       - hasDocuments:
-          count: 11
+          count: 12
 
-  - it: Should have manager service account in release namespace
+  - it: Should have manager the release namespace resource
```
Okhelifi, line 24:
```suggestion
  - it: Should have the manager namespace resource
```

make/helm.mk
```diff
@@ -4,11 +4,12 @@ KUSTOMIZE_ARGS ?= ""
...
+helm-prepare: manifests kustomize ## prepare helm chart from kustomize resources
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default -o helm/gko/templates/bundle.yaml 
 	npx zx scripts/helm-transform.mjs
 
 .PHONY: helm-template
-helm-template: manifests kustomize helm-prepare ## generates a templated bundle.yaml file in helm chart
-	helm template --include-crds helm/gko -n gko-system > bundle.yaml
+helm-template: manifests kustomize helm-prepare ## generates a templated bundle.yml
+	helm template --include-crds  helm/gko -n gko-system > bundle.yml
```
Okhelifi, line 14: we are using .yaml everywhere. Wouldn't it be better to keep .yaml and update the [doc](https://docs.gravitee.io/apim/3.x/apim_kubernetes_operator_installation_cluster.html#step_1_run_the_deployment_script) ?

release.config.js
```diff
@@ -36,6 +36,12 @@ const plugins = [
         "npx zx scripts/helm-release.mjs --version ${nextRelease.version} --img graviteeio/helm-operator",
     },
   ],
+  [
+    "@semantic-release/github",
+    {
+      assets: [{ path: "bundle.yml", label: "Operator resources bundle" }],
```
Okhelifi, line 44: If we change the file extension to yaml we will need to change it too.

Okhelifi (approved):

PR #274, opened 2023-02-21
feat: handle ingress resources with multiple hosts

a-cordier:


controllers/apim/ingress/internal/mapper/mapper.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/gravitee-io/gravitee-kubernetes-operator/pkg/keys"
+	v1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	proxyName              = "default"
+	serviceURIPattern      = "http://%s.%s.svc.cluster.local:%d"
+	flowNamePattern        = "%s%s"
```
Okhelifi, line 35: I think it's not used. We can use it on line 161 or remove it 🤷🏼

controllers/apim/ingress/internal/mapper/mapper.go
```diff
@@ -0,0 +1,338 @@
...
+		index:           index,
+	}
+}
+
+// Map maps an ingress to a graviteeio API definition, adding one virtual host per ingress rule,
+// one endpoint per backend service, and one conditional flow per host and path of the rule.
+// The host header is used to select the flow, and a dynamic routing policy routes the request
+// to the backend service, identified by the endpoint name. Is no rule matches,
+// a 404 response is returned by a flow that negates all the previous conditions.
+func (m *Mapper) Map(apiDefinition *gio.ApiDefinition, ingress *v1.Ingress) *gio.ApiDefinition {
```
Okhelifi, line 97: I was thinking if we need unit tests on this file ? 
Like discussed together it would be nice to have unit tests. With Ginkgo label we could exclude int tests and run only unit tests in the CI

Okhelifi (approved):
Tested locally 👍🏼 All good for me well done

kamiiiel (approved):

PR #281, opened 2023-02-28
Modify gateway keystore according to the ingress TLS

kamiiiel:
https://gravitee.atlassian.net/browse/APIM-626


config/samples/ingress/ingress-with-tls.yml
```diff
@@ -0,0 +1,41 @@
...
+
+# Use this context if you are running APIM and GKO in k3d
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ingress-with-tls
+  namespace: default
+  annotations:
+    kubernetes.io/ingress.class: graviteeio
+    gravitee.io/keystore: gtw-keystore-credentials
```
brasseld, line 23: Can we try to avoid that sort of thing for handling ingresses ?
In my mind, someone who's moving ingress to gravitee IC shouldn't  care specificities to ease the migration.
Happy to discuss that one @kamiiiel

config/samples/ingress/ingress-with-tls.yml
```diff
@@ -0,0 +1,41 @@
...
+  labels:
+    gravitee.io/ingress: graviteeio
```
brasseld, line 25: What's the purpose of that line ?

controllers/apim/ingress/internal/key_store.go
```diff
@@ -0,0 +1,205 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type keyStore struct {
+	name     string
+	password []byte
+}
+
+//nolint:gocognit
+func (d *Delegate) UpdateGatewayKeyStore(ingress *v1.Ingress) error {
```
brasseld, line 38: In the future, will need to make sure that the keystore is updated when one of the secret entry is itself updated (by cert-manager for example).

controllers/apim/ingress/internal/key_store.go
```diff
@@ -0,0 +1,205 @@
...
+		Namespace: ingress.Namespace,
+		Name:      secretName,
+	}
+	s := &core.Secret{}
+	if err := d.k8s.Get(d.ctx, key, s); err != nil {
+		return nil, err
+	}
+
+	return &keyStore{
+		name:     string(s.Data["name"]),
```
brasseld, line 115: IMHO, these information should probably come from the operator configuration itself, not from a secret. wdyt ?

brasseld (changes requested):

config/samples/ingress/ingress-with-tls.yml
kamiiiel: Sure, but at the end we need to know the name and the password of the GW keystore

controllers/apim/ingress/internal/key_store.go
kamiiiel: Then it needs restating the operator.

config/samples/ingress/ingress-with-tls.yml
brasseld: Agree, why I'm proposing this https://github.com/gravitee-io/gravitee-kubernetes-operator/pull/281#discussion_r1124117935

config/samples/ingress/ingress-with-tls.yml
kamiiiel: I guess this is something added by @a-cordier

controllers/apim/ingress/internal/key_store.go
brasseld: You mean, when we change the credentials ?
In that case, why not using a secret by default (something like gravitee-ic-credentials) which can be override using operator's configuration. So by default, we'll stick to the default secret, or we can let the ingress owner to override the value. wdyt ?

config/samples/ingress/ingress-with-tls.yml
brasseld: Ok, is this required ?

controllers/apim/ingress/internal/key_store.go
brasseld: I don't get this one, can we have a 📞 for discussing it ?

config/samples/ingress/ingress-with-tls.yml
kamiiiel: This seems to be only for the tests. We get all the ingresses with this label and we delete them after the tests. No other use case, it doesn't need to be added by the users.

test/ingress_controller_delete_test.go
```diff
@@ -80,6 +135,45 @@ var _ = Describe("Deleting an ingress", func() {
...
+				}
+
+				for _, a := range ks.Aliases() {
+					if a == "httpbin.example.com" {
+						return false
+					}
+				}
+
+				return true
+			}, timeout, interval).Should(BeTrue())
```
a-cordier, line 176: I find boolean assertion quite hard to understand when the CI fails. You just get a "expected false to be true" message. Maybe we should consider returning a meaning full error and `ShouldNot(HaveOccurred())` ? For the nil test I guess we could push it to the assertion.go file (AssertNotNil) to make it reusable

test/ingress_controller_create_test.go
```diff
@@ -125,5 +130,90 @@ var _ = Describe("Creating an ingress", func() {
...
+				}
+
+				for _, a := range ks.Aliases() {
+					if a == "httpbin.example.com" {
+						return true
+					}
+				}
+
+				return false
+			}, timeout, interval).Should(BeTrue())
```
a-cordier, line 215: same comment as below about the boolean assertion

controllers/apim/ingress/internal/key_store.go
```diff
@@ -0,0 +1,205 @@
...
+
+	// load exiting keystore used by the Gateway
+	ksSecret := &core.Secret{}
+	keyStore, err := d.readKeyStore(ingress, ksConfig, ksSecret)
+	if err != nil {
+		return err
+	}
+
+	updateNeed := false
+	for _, rule := range ingress.Spec.Rules {
```
a-cordier, line 58: The linter is happy because of the comment, but sonar is not https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&pullRequest=281&id=gravitee-io_gravitee-kubernetes-operator&open=AYaiVHa9uCkWqpzJ7IFU

I'm wondering if instead of silenting two flags, we should not consider a refactoring, at least extracted some of the nested logic to dedicated methods

controllers/apim/ingress/internal/key_store.go
kamiiiel: It should be fine now

controllers/apim/ingress/internal/key_store.go
kamiiiel: Now it comes from a secret that the user can specify the name or it has a default name as well

PR #283, opened 2023-03-02
feat: api definition template

kamiiiel:
https://gravitee.atlassian.net/browse/APIM-596


controllers/apim/apidefinition/apidefinition_controller.go
```diff
@@ -62,6 +63,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	delegate := internal.NewDelegate(ctx, r.Client, log)
 	events := event.NewRecorder(r.Recorder)
 
+	if apiDefinition.GetLabels()[keys.CrdApiDefinitionTemplate] == "true" {
+		log.Info("syncing template", "template", apiDefinition.Name)
+
+		if err := delegate.SyncApiDefinitionTemplate(apiDefinition, req.Namespace); err != nil {
+			log.Error(err, "Failed to sync API definition template")
+			return ctrl.Result{RequeueAfter: requeueAfterTime}, err
```
Okhelifi, line 71: When we will try to remove a template used by an ingress, we will requeue the process but in fact the user will never succeed in removing the template before the ingress. So is it really something we want to requeue ?

controllers/apim/apidefinition/internal/api_definition_template.go
```diff
@@ -0,0 +1,127 @@
...
+		if ingressList.Items[i].GetAnnotations()[keys.IngressTemplateAnnotation] == apiDefinition.Name {
+			ingresses = append(ingresses, ingressList.Items[i].GetName())
+		}
+	}
+
+	// There are existing ingresses which are still relying on this ApiDefinition template, re-schedule deletion
+	if len(ingresses) > 0 {
+		err = fmt.Errorf("can not delete %s %v depends on it", apiDefinition.Name, ingresses)
+		return err
+	}
```
Okhelifi, line 122: I think we miss an integration test on this `if` no wdyt ?

controllers/apim/apidefinition/internal/api_definition_template.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err := d.update(apiDefinition, namespace); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d *Delegate) update(apiDefinition *gio.ApiDefinition, namespace string) error {
```
Okhelifi, line 57: Is it really an update method. I mean here we are in the `api_definition_template.go` and the method is named `update`. But in reality here we are updating ingress liked to the template not the template. I don't know if it would be better to name it `updateIngresses` wdyt ?

test/internal/fixtures.go
```diff
@@ -57,6 +60,7 @@ func (f *FixtureGenerator) AddSuffix(property string) string {
 	return property + f.Suffix
 }
 
+//nolint:gocognit
```
Okhelifi, line 63: Sonar looks not happy about the complexity of this method but I think we can address it in a dedicated PR

Okhelifi (commented):
Some minor concerns but I tested it locally and it's working 👍🏼

test/internal/fixtures.go
a-cordier: I would say this can wait for another PR as it is in the fixtures file

controllers/apim/apidefinition/internal/api_definition_template.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Listing ingresses from the same namespace
+	err := d.k8s.List(d.ctx, &ingressList, client.InNamespace(namespace))
+
+	if err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	for i := range ingressList.Items {
+		ingress := &ingressList.Items[i]
```
Okhelifi, line 71:
```suggestion
        for _, ingress := range ingressList.Items {
```

kamiiiel: I did it like this to avoid copying the whole ingress in to the memory

controllers/apim/apidefinition/internal/api_definition_template.go
kamiiiel: I added a new test

test/internal/fixtures.go
```diff
@@ -18,29 +18,32 @@ import (
...
 
 var decode = scheme.Codecs.UniversalDecoder().Decode
 
 type Fixtures struct {
-	Api      *gio.ApiDefinition
-	Context  *gio.ManagementContext
-	Resource *gio.ApiResource
-	Ingress  *netV1.Ingress
+	Api         *gio.ApiDefinition
+	ApiTemplate *gio.ApiDefinition
```
Okhelifi, line 35: Is it necessary to add a dedicated fiels in the fixture as an `ApiTemplate` is an `Api` ?
i don't know if we want to keep both or just use the `internal.ApiTemplateWithApiKeyPlanFile` constant with the `Api` attribute

test/ingress_controller_delete_test.go
```diff
@@ -82,4 +84,153 @@ var _ = Describe("Deleting an ingress", func() {
...
+			}, timeout, interval).ShouldNot(HaveOccurred())
+
+			expectedApiName := ingressFixture.Name
+			Expect(createdApiDefinition.Name).Should(Equal(expectedApiName))
+
+			Expect(len(createdApiDefinition.Spec.Plans)).Should(Equal(1))
+			Expect(createdApiDefinition.Spec.Plans[0].Security).Should(Equal("API_KEY"))
+		})
+
+		AfterEach(func() {
```
Okhelifi, line 154: We discussed about the AfterEach in a previous PR and decided it's not really necessary as everything is removed in the AfterEach of the suite test file.
Maybe we can remove it 🤷🏼 

( see https://github.com/gravitee-io/gravitee-kubernetes-operator/pull/237#discussion_r1102828551 )

test/ingress_controller_update_test.go
```diff
@@ -96,4 +100,160 @@ var _ = Describe("Updating an ingress", func() {
...
+
+			ingressFixture = fixtures.Ingress
+			ingressFixture.Annotations[keys.IngressTemplateAnnotation] = apiDefinitionTemplate.Name
+			ingressLookupKey = types.NamespacedName{Name: ingressFixture.Name, Namespace: namespace}
+
+			By("Creating an Ingress and the default ApiDefinition")
+			Expect(k8sClient.Create(ctx, ingressFixture)).Should(Succeed())
+		})
+
+		AfterEach(func() {
```
Okhelifi, line 139: Same here for the AfterEach 🤷🏼

PR #317, opened 2023-04-13
refactor: improve error messages

a-cordier:


Okhelifi (approved):

Okhelifi:
There is a code smell about the "gravitee.yml" we can correct I think except that it's ok

PR #333, opened 2023-05-04
feat: application CRD

kamiiiel:
**Issue:**
https://gravitee.atlassian.net/browse/APIM-767


controllers/apim/managementcontext/internal/delete.go
```diff
@@ -52,6 +52,22 @@ func Delete(
...
+		app,
+	)
+
+	if err != nil {
+		err = fmt.Errorf("an error occurred while checking if the management context is linked to an application: %w", err)
+		return err
+	}
+
+	if len(app.Items) > 0 {
+		return fmt.Errorf("can not delete %s because %d api(s) relying on this context", instance.Name, len(app.Items))
```
a-cordier, line 68:
```suggestion
		return fmt.Errorf("can not delete %s because %d application(s) are relying on this context", instance.Name, len(app.Items))
```

a-cordier (changes requested):

a-cordier (approved):

PR #349, opened 2023-05-12
feat: template resolver

kamiiiel:
https://gravitee.atlassian.net/browse/APIM-1584

### DIFF
--- a/internal/env/template/resolver.go
+++ b/internal/env/template/resolver.go
@@ -1,0 +1,169 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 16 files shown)


internal/env/template/resolver.go
```diff
@@ -0,0 +1,169 @@
...
+	obj    runtime.Object
+}
+
+func NewResolver(ctx context.Context, c client.Client, l logr.Logger, obj runtime.Object) *Resolver {
+	return &Resolver{ctx: ctx, client: c, log: l, obj: obj}
+}
+
+func (r *Resolver) Resolve() error {
+	switch t := r.obj.(type) {
+	case *gio.ApiDefinition, *gio.ManagementContext, *gio.Application, *netv1.Ingress:
```
a-cordier, line 55: Given the change added to the resource controller, I think we are missing the ApiResource object type here.

a-cordier (changes requested):

a-cordier (approved):

PR #387, opened 2023-06-08
docs: update release section

a-cordier:


julian-robotham-gravitee (approved):
Reviewed the wording and the process appears to be clear from my stand point.

PR #415, opened 2023-08-11
feat: handle v4 definition (part 1)

a-cordier:
As we will need to implement webhook conversion for v4 api model storage, the idea here is to bootstrap the v4 model and to implement basic conversion.


api/model/api/v4/listeners.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type Listener struct {
+	BaseListener         `json:",inline"`
+	HttpListener         *HttpListener         `json:",inline"`
+	TCPListener          *TCPListener          `json:",inline"`
+	SubscriptionListener *SubscriptionListener `json:",inline"`
+}
+
+func (l *Listener) UnmarshalJSON(data []byte) error {
```
a-cordier, line 114: I don't know if we can get around this, the Listener type in the JSON payload is polymorphic and I did not find another way than overriding the unmarshal method 🤷‍♂️

kamiiiel (approved):

PR #426, opened 2023-09-05
fix: set default policy execution mode to v4-emulation-engine

a-cordier:
see https://gravitee.atlassian.net/browse/APIM-2498

Not setting execution_mode to `v4-emulation-engine` in our model makes execution policy behaviour inconsistent between an API created from the operator and the same API created from the UI.

Tested on 3.20.x and setting this field with this value falls back to v3 there. So that should not introduce any backward compatibility issue.

Waiting confirmation from ENR before making this ready to review.


a-cordier:
> Waiting confirmation from ENR before making this ready to review.

A fix is coming on APIM as well

PR #438, opened 2023-09-20
feat: allow install without any cluster role

a-cordier:
Some constrained env do not allow to use cluster role. The idea here is to
  - not create a cluster role for CRDs if `.Values.manager.applyCRDs` is set to false
  - not create a cluster role for the metric server if `.Values.manager.metrics.enabled` is set to false
  - disable cluster role for the RBAC proxy if metrics are disabled (It is not used in that case)
  - allow to disable the RBAC proxy
  - allow to skip RBAC creation
  - allow to create RBAC but to skip cluster roles
  - allow to use custom image / tag for the RBAC proxy
  - allow to chose a service account name
  - allow to skip service account creation
  
To make this possible, we move from the kustomize -> helm process to plain helm templating

the config folder has been cleaned up a bit, but I kept some of the files for the time being (to be continued)


helm/gko/templates/manager/deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+        control-plane: controller-manager
+    spec:
+      containers:
+{{- if and .Values.manager.metrics.enabled .Values.rbacProxy.enabled }}
+        - args:
+            - --secure-listen-address=0.0.0.0:8443
+            - --upstream=http://127.0.0.1:8080/
+            - --logtostderr=true
+            - --v=0
+          image: gcr.io/kubebuilder/kube-rbac-proxy:v0.11.0
```
kamiiiel, line 46: This needs to be updated with the new version we applied already

kamiiiel (approved):

PR #442, opened 2023-09-26
fix: unmarshal int values in GenericMap

kamiiiel:
https://github.com/gravitee-io/issues/issues/9260


a-cordier:
👍 Could we add an IT test just with the spike arrest descriptor (one that just ensure the operator does not panic when we apply the definition and don't test all the way down to the gateway ?

a-cordier (commented):
👍 Could we add an integration test ?

One that just ensures the manager does not panic when we try to apply a definition with the spike arrest config given by the OP ? I don't think we need to test this all the way up to the gateway.

a-cordier (approved):

kwinkel:
Any chance to publish this change? We are currently facing `Observed a panic in reconciler: cannot deep copy int` and would like to give it a shot if this fixes it.

PR #456, opened 2023-10-07
docs(helm): use bitnami readme generator

a-cordier:
Bitnami readme generator allow to organise Helm documentation into sections, which makes the document somehow more readable.

⚠️ To keep sections consistants, the `httpClient` value has been moved to the manager section.

Which means this should be released in 1.0.0 only and documented as a breaking change.

### DIFF
--- a/helm/gko/templates/manager/config.yaml
+++ b/helm/gko/templates/manager/config.yaml
@@ -45,6 +45,6 @@ data:
   {{- if $template404.namespace }}
   TEMPLATE_404_CONFIG_MAP_NAMESPACE: {{ $template404.namespace }}
   {{- end }}
-  {{- if .Values.httpClient.insecureSkipCertVerify }}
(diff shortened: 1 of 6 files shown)


helm/gko/templates/manager/config.yaml
```diff
@@ -37,6 +37,6 @@ data:
   {{- if $template404.namespace }}
   TEMPLATE_404_CONFIG_MAP_NAMESPACE: {{ $template404.namespace }}
   {{- end }}
-  {{- if .Values.httpClient.insecureSkipCertVerify }}
-  INSECURE_SKIP_CERT_VERIFY: {{ .Values.httpClient.insecureSkipCertVerify }}
+  {{- if .Values.manager.httpClient.insecureSkipCertVerify }}
+  INSECURE_SKIP_CERT_VERIFY: "true"
```
kamiiiel, line 41: the default value should be "false"

a-cordier: Yes, that what it is doing here

kamiiiel (approved):

PR #457, opened 2023-10-08
docs: improve crdoc output

a-cordier:


api/model/api/v2/services.go
```diff
@@ -20,22 +20,36 @@ import (
...
+	*Service `json:",inline"`
+
+	// Provider name
+	Provider string `json:"provider,omitempty"`
+
+	// Configuration, arbitrary map of key-values
+	Config *utils.GenericStringMap `json:"configuration,omitempty"`
+
+	// Is it secondary or not?
+	Secondary bool `json:"secondary,omitempty"`
```
a-cordier, line 49: That's weird, I don't think this property is defined in the definition model of the GW for endpoint discovery services.

api/model/api/v4/analytics.go
```diff
@@ -15,15 +15,27 @@
...
+	// Logging mode
+	Mode *LoggingMode `json:"mode,omitempty"`
+
+	// Logging phase
+	Phase *LoggingPhase `json:"phase,omitempty"`
 }
 
 type LoggingPhase struct {
-	Request  bool `json:"request"`
+	// Is logging phase a Request or not?
```
a-cordier, line 35:
```suggestion
	// Should the request phase of the request roundtrip be included in the log payload or not ?
```

PR #497, opened 2023-11-06
feat: introduce pem registry

kamiiiel:
https://gravitee.atlassian.net/browse/APIM-2969


a-cordier:
While testing the PR, I had to make a change on gravitee-node to get the init of the keystore loader non blocking. meaning changing `blockingAwait` to `subscribe` [here](https://github.com/gravitee-io/gravitee-node/blob/821699e5bf7c6839e4383aa7da461355856af60f/gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/AbstractKubernetesKeyStoreLoader.java#L72) 

The keystore gets updated on gateway side, under the condition that the `cert-manager.io/common-name` annotation is populated on the ingress (other wise the reconcile fails, because the key ends up empty in the registry.

When using a wild card in the Common Name of the certificate, the reconcile fails because `*` is an illegal character in a config map key.

a-cordier (commented):
While testing the PR, I had to make a change on gravitee-node to get the init of the keystore loader non blocking. meaning changing `blockingAwait` to `subscribe` [here](https://github.com/gravitee-io/gravitee-node/blob/821699e5bf7c6839e4383aa7da461355856af60f/gravitee-node-kubernetes/src/main/java/io/gravitee/node/kubernetes/keystoreloader/AbstractKubernetesKeyStoreLoader.java#L72) 

The keystore gets updated on gateway side, under the condition that the `cert-manager.io/common-name` annotation is populated on the ingress (other wise the reconcile fails, because the key ends up empty in the registry.

When using a wild card in the Common Name of the certificate, the reconcile fails because `*` is an illegal character in a config map key.

internal/indexer/indexer.go
```diff
@@ -120,3 +121,17 @@ func IndexApplicationManagementContexts(application *gio.Application, fields *[]
 
 	*fields = append(*fields, application.Spec.Context.String())
 }
+
+func IsGraviteeIngress(ingress *v1.Ingress) bool {
```
a-cordier, line 125: maybe we could make this utility function part of an ingress.go file in the k8s package ? It's not ideal, but `k8s.IsGraviteeIngress` sounds more logical than `indexer.IsGraviteeIngress` to me WDYT ?

a-cordier (approved):

PR #553, opened 2024-01-26
fix: use empty arrays instead of null in serialized JSON

kamiiiel:


a-cordier (commented):
Just a thought, could we not just remove `omitempty` on slice properties and use `kubebuilder:default:={}` as described here ? https://book.kubebuilder.io/reference/markers/crd-validation

a-cordier (approved):

PR #578, opened 2024-02-26
fix: make member display name and email optional

a-cordier:


api/model/api/base/member.go
```diff
@@ -15,8 +15,12 @@
 package base
 
 type Member struct {
-	Id          string `json:"id"`
-	Email       string `json:"email"`
-	DisplayName string `json:"displayName"`
-	Type        string `json:"type"`
+	// +kubebuilder:validation:Optional
```
kamiiiel, line 18: These are all Oprional? @a-cordier 
So what happens if they send all empty?

kamiiiel (approved):

PR #660, opened 2024-06-03
feat: add documentation page support for V4 APIs

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-32


api/model/api/base/api.go
```diff
@@ -62,6 +62,13 @@ type ApiBase struct {
 	// The list of categories the API belongs to.
 	// Categories are reflected in APIM portal so that consumers can easily find the APIs they need.
 	Categories []string `json:"categories"`
+	// +kubebuilder:validation:Optional
+	// A map of unique page identifiers to pages
+	// Keys uniquely identify pages and are used to keep them in sync
+	// with APIM when using a management context. This means that renaming
+	// a key is same as deleting the previous page associated to that key,
+	// and generating a new one, holding a new ID in APIM.
```
benoitgravitee, line 70:
```suggestion
	// +kubebuilder:validation:Optional
	// A map of pages objects.
	// 
	// Keys uniquely identify pages and are used to keep them in sync
	// with APIM when using a management context. 
	//  
	// Renaming a key is the equivalent of deleting the page and recreating 
	// it holding a new ID in APIM.
```

internal/apim/model/page.go
```diff
@@ -15,17 +15,19 @@
...
 )
 
 type Page struct {
-	ID      string `json:"id,omitempty"`
-	Content string `json:"content,omitempty"`
-	Type    string `json:"type,omitempty"`
+	ID      string                 `json:"id,omitempty"`
+	Content string                 `json:"content,omitempty"`
+	Type    string                 `json:"type,omitempty"`
+	Source  map[string]interface{} `json:"source,omitempty"`
```
benoitgravitee, line 26: Out of curiosity, why don't you create a type for Source with `Configuration: map[string]interface{}` and `Type: string` ?

test/integration/apidefinition/v4_create_withContext_andSwaggerHTTPFetcher_test.go
```diff
@@ -0,0 +1,83 @@
...
+				page = pages[0]
+			} else {
+				page = pages[1]
+			}
+
+			if err = assert.Equals("fetcher type", page.Source.Type, swagger.Source.Type); err != nil {
+				return err
+			}
+
+			return assert.Equals("fetcher source", page.Source.Configuration, swagger.Source.Configuration)
```
a-cordier, line 80: I think we should test that the content is not empty here. This will actually fail as something is missing in APIM to get there

a-cordier (commented):
I think updates are required to https://github.com/gravitee-io/gravitee-api-management/pull/7648 so that we can go on with this one

test/integration/apidefinition/v4_create_withContext_andSwaggerHTTPFetcher_test.go
a-cordier: should be feasible when https://github.com/gravitee-io/gravitee-api-management/pull/7648 is merged

a-cordier (approved):

PR #664, opened 2024-06-10
ci: update release job and drop semantic release

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-15

Done
--

The pipeline as the following parameters:

```yaml
  trigger:
    type: enum
    default: test
    enum:
      - release
      - test
  release-version:
    type: string
    default: ""
  latest:
    type: boolean
    default: false
    description: Is this version the latest version available ?
  dry-run:
    type: boolean
    default: true
```

  - trigger should alway be release in that case
  - if latest is true, both the docker image and the github release will be flagged as latest
  - if dry run is set to false the additional steps will be 
    - pushing the docker image to docker hub with latest tag (as this is the latest version)
    - release the helm chart on https://helm.gravitee.io/
    - push release commit and tag to the release branch (4.4.X)
    - publish the github release with the changelog we generate from Jira and the custom 
    resource definitions packaged as a zip file
    - announce the release on the release alert channel

The github release Changelog is generated from the tickets found for the version in our Jira backlog

The trigger-release github action allows to call the pipeline from the repo itself and looks like this

Screenshot 2024-06-11 at 16 19 31

To Be Done
--
Submit changelog / release note as a PR for tech writers


.circleci/config.yml
```diff
@@ -7,25 +7,34 @@ orbs:
...
+  release-version:
+    type: string
+    default: ""
+  latest:
+    type: boolean
+    default: false
+    description: Is this version the latest version available ?
+  dry-run:
+    type: boolean
+    default: false
```
benoitgravitee, line 31: Maybe true by default to avoid unwanted release.

.circleci/config.yml
```diff
@@ -56,19 +65,27 @@ commands:
...
                 name: Setting docker image tag
                 command: echo "export DOCKER_IMAGE_TAG=master-latest" >> ${BASH_ENV}
       - when:
           condition:
-            equal: [gravitee_gko_dockerhub_integration_id, << parameters.integration_id_key >>]
+            equal:
+              [
+                gravitee_gko_dockerhub_integration_id,
+                << parameters.integration_id_key >>,
+              ]
```
benoitgravitee, line 86:
```suggestion
            equal:
                - gravitee_gko_dockerhub_integration_id
                - << parameters.integration_id_key >>
```

.circleci/config.yml
```diff
@@ -341,39 +357,153 @@ jobs:
...
+            git diff HEAD~1 HEAD
+      - when:
+          condition:
+            not: << pipeline.parameters.dry-run >>
+          steps:
+            - run:
+                name: Push release tag and commit
+                command: |
+                  export RELEASE_BRANCH=$(npx zx scripts/get-release-branch.mjs --version << pipeline.parameters.release-version >>)
+                  git push -u origin $RELEASE_BRANCH
```
benoitgravitee, line 467: a detail: -u is probably not required here.

.github/workflows/trigger-release.yml
```diff
@@ -14,21 +14,48 @@
...
+        default: true
+        type: boolean
+      pipeline-branch:
+        description: 'On which branch is your circleci pipeline located ?'
+        required: true
+        default: "master"
+        type: string
 jobs:
   trigger-release:
     runs-on: ubuntu-latest
```
benoitgravitee, line 41: maybe there is leaner distribution to run things on ?

scripts/run-kind.mjs
```diff
@@ -14,82 +14,80 @@
  * limitations under the License.
  */
 
-import { LOG, setNoQuoteEscape, setQuoteEscape, time, toggleVerbosity } from "./lib/index.mjs";
```
benoitgravitee, line 17: Just linting ? Hard to see what changed

a-cordier: Yes it's because I could not help but run prettier in the script dir (and on the circleci config file) :)

.github/workflows/trigger-release.yml
a-cordier: I tried to have a look but apparently leaner options are for private repositories https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job

benoitgravitee (approved):

PR #665, opened 2024-06-10
refactor: rename application CRD's "applicationMetadata" and "applica…

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-4

### DIFF
--- a/api/model/application/application.go
+++ b/api/model/application/application.go
@@ -15,8 +15,6 @@
 // +kubebuilder:object:generate=true
 package application
 
-import "github.com/gravitee-io/gravitee-kubernetes-operator/api/model/utils"
-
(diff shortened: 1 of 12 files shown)


api/model/application/application.go
```diff
@@ -113,5 +105,5 @@ type Application struct {
 	// Should membership notifications be disabled or not ?
 	DisableMembershipNotifications bool `json:"disable_membership_notifications,omitempty"`
 	// Application meta data
-	ApplicationMetaData *[]MetaData `json:"applicationMetaData,omitempty"`
```
a-cordier, line 116: From the issue description, I think we should keep this and mark it as deprecated for BWC

a-cordier (approved):

PR #703, opened 2024-06-27
Update README.adoc

jmcx:
fix broken doc links, since doc updates


jmcx:
remade a new PR with better commit header

PR #707, opened 2024-06-28
refactor: sync using new application crd endpoint

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-276

### DIFF
--- a/api/model/application/application.go
+++ b/api/model/application/application.go
@@ -16,7 +16,7 @@
 package application
 
 // +kubebuilder:validation:Enum=SHARED;EXCLUSIVE;UNSPECIFIED;
-type KeyMode string
+type AppKeyMode string
 
 type SimpleSettings struct {
 	// Application Type
@@ -76,7 +76,7 @@ type Application struct {
 	// Application Description
 	Description string `json:"description,omitempty"`
 	// Application Type
-	ApplicationType string `json:"type,omitempty"`
+	Type string `json:"type,omitempty"`
 	// The ClientId identifying the application. This field is required when subscribing to an OAUTH2 / JWT plan.
 	ClientId string `json:"clientId,omitempty"`
 	// List of application Redirect Uris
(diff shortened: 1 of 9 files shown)


api/model/application/application.go
```diff
@@ -99,11 +99,16 @@ type Application struct {
...
 	// Application settings
-	Settings *Setting `json:"settings,omitempty"`
+	// +kubebuilder:validation:Required
+	Settings *Setting `json:"settings"`
 	// The API key mode to use. If shared, the application will reuse the same API key across various subscriptions.
-	AppKeyMode *KeyMode `json:"app_key_mode,omitempty"`
+	AppKeyMode *AppKeyMode `json:"app_key_mode,omitempty"`
 	// Should membership notifications be disabled or not ?
 	DisableMembershipNotifications bool `json:"disable_membership_notifications,omitempty"`
+	// If the app is ACTIVE or ARCHIVED.
```
a-cordier, line 108: actually ARCHIVED only stands for deleted app. I wonder if it should even be there ... WDYT ?

internal/apim/service/applications.go
```diff
@@ -81,23 +83,15 @@ func (svc *Applications) GetMetadataByApplicationID(appId string) (*[]model.Appl
 	return application, nil
 }
 
-func (svc *Applications) CreateUpdate(method string, spec *application.Application) (*model.Application, error) {
-	url := svc.EnvV1Target(applicationsPath)
-	if spec.ID != "" {
-		url = svc.EnvV1Target(applicationsPath).WithPath(spec.ID)
-	}
+func (svc *Applications) CreateUpdate(spec *application.Application) (*v1alpha1.ApplicationStatus, error) {
```
a-cordier, line 86: maybe we could rename this to `CreateOrUpdate` WDYT ?

a-cordier (approved):

PR #721, opened 2024-07-08
feat: add validation webhook to GKO

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-287


main.go
```diff
@@ -282,3 +301,40 @@ func applyCRDs() error {
 		return err
 	})
 }
+
+func pathAdmissionWebhook() {
```
a-cordier, line 305: not sure but I guess this function is intended to be called `patchAdmissionWebhook ` am I wrong ?

internal/certs/certs_test.go
```diff
@@ -0,0 +1,77 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
```
a-cordier, line 1: could we move this file to the unit directory ?

This way we have a much more simple way to run IT tests and UT in different CI jobs (currently I don't think this one is run in our pipeline)

api/v1alpha1/apiv4definition_webhook.go
```diff
@@ -0,0 +1,64 @@
...
+func (api *ApiV4Definition) ValidateCreate() (admission.Warnings, error) {
+	// make sure Management Context exist before creating the API Definition resource
+	if api.HasContext() {
+		cxtRef := api.Spec.Context
+		mCxt := new(ManagementContext)
+		err := k8s.GetClient().Get(context.Background(), cxtRef.NamespacedName(), mCxt)
+		if errors.IsNotFound(err) {
+			return nil, errors.NewUnrecoverableError(
+				fmt.Sprintf("can't create api [%s] because it is using "+
+					"management context [%s] that doesn't exist in the cluster", api.Name, cxtRef.Name))
```
a-cordier, line 51: Maybe we should use `cxtRef.NamespacedName()` in the error message as the context might site in another NS

internal/webhook/certs.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+const (
+	CertName = "cert"
+	KeyName  = "key"
+	caName   = "ca"
+	Name     = "gko-validating-webhook-configurations"
+)
+
+type Pather struct {
```
a-cordier, line 45:
```suggestion
type Patcher struct {
```

a-cordier (approved):

PR #735, opened 2024-07-16
feat: add validation webhook to GKO

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-321

### DIFF
--- a/api/v1alpha1/managementcontext_webhook.go
+++ b/api/v1alpha1/managementcontext_webhook.go
@@ -18,12 +18,15 @@ package v1alpha1
 
 import (
 	"context"
+	"fmt"
 	"net"
 
 	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim/client"
 	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/http"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/k8s"
 	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/uuid"
-	errors "github.com/pkg/errors"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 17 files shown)


api/v1alpha1/managementcontext_webhook.go
```diff
@@ -45,3 +56,23 @@ func (context *ManagementContext) ValidateUpdate(_ runtime.Object) (admission.Wa
...
+	urLs, _ := client.NewURLs(ctx.Spec.BaseUrl, ctx.Spec.OrgId, ctx.Spec.EnvId)
+
+	httpClient := http.NewClient(context.Background(), nil)
+	cli := client.Client{
+		HTTP: httpClient,
+		URLs: urLs,
+	}
+
+	api := make(map[string]interface{})
+	err := httpClient.Get(cli.EnvV1Target("apis").WithPath(uuid.NewV4String()).String(), api)
```
a-cordier, line 70: Why not add a Get() function in the [apim.Env](https://github.com/gravitee-io/gravitee-kubernetes-operator/blob/master/internal/apim/service/env.go) service ? It would make the code a bit simpler and this way we could as well check that the environment referenced in the context exists WDYT ?

a-cordier: that's why we should have an admission package in internal, and not embedded validation as a behaviour of the object it self no ?

a-cordier: Let's move forward and refactor later :)

api/v1alpha1/managementcontext_webhook.go
```diff
@@ -35,6 +42,10 @@ func (context *ManagementContext) SetupWebhookWithManager(mgr ctrl.Manager) erro
 func (context *ManagementContext) Default() {}
 
 func (context *ManagementContext) ValidateCreate() (admission.Warnings, error) {
+	if err := resolveManagementContext(context); err != nil {
+		return admission.Warnings{}, err
```
a-cordier, line 46: We should actually return a warning on connection errors (CF issue description)

a-cordier (changes requested):

a-cordier (approved):

PR #752, opened 2024-07-22
refactor: look for existing APIs in the same namespace only

kamiiiel:
https://gravitee.atlassian.net/browse/GKO-434


api/v1alpha1/apiv4definition_webhook.go
```diff
@@ -64,7 +66,9 @@ func validateApi(api *ApiV4Definition) (admission.Warnings, error) {
 	} else {
 		// check for unique context path
 		apis := new(ApiV4DefinitionList)
-		if err := k8s.GetClient().List(context.Background(), apis); err != nil {
+		if err := k8s.GetClient().List(context.Background(), apis, &client.ListOptions{
+			Namespace: api.Namespace,
```
a-cordier, line 70: would we have a problem if the gateway and the operator are configured using a cluster role, and therefore listening both to any namespace ?

a-cordier (approved):

PR #776, opened 2024-08-07
refactor: revert back previous changes to API resource fields

kamiiiel:
https://gravitee.atlassian.net/browse/GKO-513

### DIFF
--- a/api/model/api/base/resource.go
+++ b/api/model/api/base/resource.go
@@ -28,11 +28,11 @@ type Resource struct {
 	// Is resource enabled or not?
 	Enabled bool `json:"enabled"`
 	// Resource Name
-	Name string `json:"name,omitempty"`
+	Name string `json:"name"`
 	// Resource Type
-	Type string `json:"type,omitempty"`
+	Type string `json:"type"`
 	// Resource Configuration, arbitrary map of key-values
-	Configuration *utils.GenericStringMap `json:"configuration,omitempty"`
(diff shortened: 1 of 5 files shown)


api/model/api/base/resource.go
```diff
@@ -28,11 +28,11 @@ type Resource struct {
...
+	Name string `json:"name"`
 	// Resource Type
-	Type string `json:"type,omitempty"`
+	Type string `json:"type"`
 	// Resource Configuration, arbitrary map of key-values
-	Configuration *utils.GenericStringMap `json:"configuration,omitempty"`
+	Configuration *utils.GenericStringMap `json:"configuration"`
 }
 
 func (r *Resource) GetType() string {
```
a-cordier, line 38: The problem is not there 

Please remove this and have a look at the getters in this file, one is wrong

```
func (r *Resource) GetType() string {
-       return r.Name
+       return r.Type
}
```

Also, in the admission controller, I forgot to call validateCreate with the new object in the update method (hence the bug on updates)

Sorry about that 🙏

a-cordier (changes requested):

api/model/api/base/resource.go
a-cordier: As a side note, I think we should improve the error message to emphasise that the missing property is in a resource

something like 

`"missing required value in API resource property [type]"`

a-cordier: Also, please rename the commit as it's not a revert anymore :)

PR #778, opened 2024-08-08
fix: resolve resource ref while validating the resources

kamiiiel:
https://gravitee.atlassian.net/browse/GKO-513


a-cordier:
I don't know about the approach TBH, this can be fixed with two lines of diff (except for tests). It's really just about the wrong getter I mentioned in the closed PR and the validateUpdate that is not there. Plus, I don't see the missing update problem fixed here.

internal/admission/api/base/resource.go
```diff
@@ -25,12 +29,36 @@ import (
...
+					Namespace: res.GetRef().GetNamespace(),
+				})
+			}
 		} else {
+			references[i] = toResourceOrRef(res.GetObject(), nil)
 			errs.MergeWith(resource.ValidateModel(ctx, res.GetObject()))
 		}
 	}
+
+	api.SetResources(references)
```
benoitgravitee, line 50: Why doing this all the time ? Is it case we implement this for v2 ?

internal/admission/api/base/resource.go
```diff
@@ -25,12 +29,36 @@ import (
...
 	errs := errors.NewAdmissionErrors()
-	for _, res := range api.GetResources() {
-		if res.IsRef() && dynamic.ExpectResolvedResource(ctx, res.GetRef(), api.GetNamespace()) != nil {
-			errs.AddSevere("api references resource [%s] that does not exist in the cluster", res.GetRef())
+	references := make([]core.ObjectOrRef[core.ResourceModel], len(api.GetResources()))
+	for i, res := range api.GetResources() {
+		if res.IsRef() {
+			if r, err := dynamic.ResolveResource(ctx, res.GetRef(), api.GetNamespace()); err != nil {
+				errs.AddSevere("api references resource [%s] that does not exist in the cluster", res.GetRef())
+				references[i] = res
```
benoitgravitee, line 37: You probably don't need to do this.

benoitgravitee (approved):

internal/admission/api/base/resource.go
a-cordier: I would maybe have gone with a SetObject method on the ObjectOrRef interface WDYT ?

I think it would remove a bunch of else statements here

PR #791, opened 2024-08-19
feat: add member support in Application CRD

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-381


api/model/application/application.go
```diff
@@ -53,6 +53,17 @@ type Metadata struct {
...
 
+type Member struct {
+	// User's technical identifier
+	ID string `json:"id"`
+	// User's reference for user providing from an identity provider
+	Reference *string `json:"reference,omitempty"`
+	// Role's name
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=`USER`
+	Role string `json:"role"`
```
a-cordier, line 64: Is there a way we could make API member and Application member looking the same ?

a-cordier (approved):

PR #826, opened 2024-09-19
feat: add validation for templating

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-422

### DIFF
--- a/internal/admission/api/base/validate.go
+++ b/internal/admission/api/base/validate.go
@@ -17,6 +17,8 @@ package base
 import (
 	"context"
 
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/admission"
+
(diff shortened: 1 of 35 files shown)


internal/admission/api/base/validate.go
```diff
@@ -26,6 +28,8 @@ import (
 func ValidateCreate(ctx context.Context, obj runtime.Object) *errors.AdmissionErrors {
 	errs := errors.NewAdmissionErrors()
 
+	errs.Add(admission.ValidateTemplate(ctx, obj))
```
a-cordier, line 31: should we compile the template before validating the resources ? I don't think what we have here is the spec that we will apply eventually

benoitgravitee (commented):
looks good enough for me, a few remarks on typos and code clarity but nothing major

a-cordier (approved):

PR #854, opened 2024-10-05
fix: keep default values for all properties

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-654


api/model/api/base/resource.go
```diff
@@ -25,22 +25,26 @@ var _ core.ObjectOrRef[core.ResourceModel] = &ResourceOrRef{}
...
+	// +kubebuilder:validation:Optional
+	Type *string `json:"type,omitempty"`
 	// Resource Configuration, arbitrary map of key-values
+	// +kubebuilder:validation:Optional
 	Configuration *utils.GenericStringMap `json:"configuration,omitempty"`
 }
 
 func (r *Resource) GetType() string {
-	return r.Type
+	return *r.Type
```
a-cordier, line 43: I'm not even sure it can happen, but should we add a nil check to avoid panicking at runtime ?

a-cordier (approved):

PR #882, opened 2024-10-16
adding tests for groups in v2 APIs

ctschacher:
Adding tests for groups in v2 APIs. 

- Add 1 group 
- add multiple groups 
- delete group

https://gravitee.atlassian.net/browse/GKO-716

### DIFF
--- a/test/integration/apidefinition/v2/create_withContext_andMultipleGroups_test.go
+++ b/test/integration/apidefinition/v2/create_withContext_andMultipleGroups_test.go
@@ -1,0 +1,93 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//         http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v2
+
+import (
+	"context"
+	"strings"
+
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim/model"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/apim"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/assert"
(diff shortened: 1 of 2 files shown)


test/integration/apidefinition/v2/create_withContext_andMultipleGroups_test.go
```diff
@@ -0,0 +1,74 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Create", labels.WithContext, func() {
+	timeout := constants.EventualTimeout
+	interval := constants.Interval
+
+	ctx := context.Background()
+
+	It("should create one group", func() {
```
a-cordier, line 39: I guess this should be updated to reflect the fact that there are more than one group ?

Also, we don't create a group here, do we ? The operator simply attaches the groups to the API, maybe we should think about another title, especially given that group creation with the operator might become an actual thing.

a-cordier: As an addition, that's a good idea to attach several groups at once, but how about melting this in another test about group deletion ?

the test would create 2 or 3 groups and remove one of them, then assert that APIM has been updated as it should.

WDYT ?

test/integration/apidefinition/v2/create_withContext_andMultipleGroups_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+		Eventually(func() error {
+			apiExport, err := apim.Export.V2Api(fixtures.API.Status.ID)
+			if err != nil {
+				return err
+			}
+
+			exportedGroups := apiExport.Spec.Groups
+			sort.Strings(exportedGroups)
+			return assert.Equals("groups", expectedGroups, exportedGroups)
```
a-cordier, line 71: You can use
```golang
assert.SliceEqualsSorted("groups", expectedGroups, exportedGroups)
```

test/integration/apidefinition/v2/create_withContext_andOneGroup_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		Eventually(func() error {
+			apiExport, err := apim.Export.V2Api(fixtures.API.Status.ID)
+			if err != nil {
+				return err
+			}
+
+			exportedGroups := apiExport.Spec.Groups
+
+			return assert.Equals("groups", expectedGroups, exportedGroups)
```
a-cordier, line 65: Why don't we sort there ?

test/integration/apidefinition/v2/create_withContext_andMultipleGroups_test.go
ctschacher: Oh, nice one, thanks!

a-cordier (approved):
👍

PR #885, opened 2024-10-19
test: add category tests for v2 APIs

ctschacher:
### Integration Tests:
Added integration test to verify the creation of an API 
* with a single category
* with multiple categories and the ability to remove a category (update)

https://gravitee.atlassian.net/browse/GKO-720


internal/apim/model/group.go
```diff
@@ -18,3 +18,8 @@ type Group struct {
 	ID   string `json:"id,omitempty"`
 	Name string `json:"name,omitempty"`
 }
+
+type Category struct {
```
a-cordier, line 22: maybe we should rename the file now that we have several models defined in it WDYT ?

Maybe `env.go` would be a god fit ?

a-cordier (commented):
just a minor comment otherwise 👍

internal/apim/model/group.go
ctschacher: Ok, good. I'll merge the content of the `group.go` into `env.go`.

PR #928, opened 2024-11-26
fix: add API failover to the 4 API definitions

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-784


a-cordier (commented):
I think we need another commit on APIM regarding the export, because when I tried to test this, enabling failover on an API an exporting it does not include the object in the resource.

kamiiiel:
> I think we need another commit on APIM regarding the export, because when I tried to test this, enabling failover on an API an exporting it does not include the object in the resource.

Will fix it in the APIM export side

a-cordier (approved):

PR #931, opened 2024-11-27
fix: 2 or more API plans of the same type will be create in APIM

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-804


a-cordier (approved):
could we rephrase the commit to something like 
`Unable to import two v2 plans with same type in APIM` ?

PR #951, opened 2024-12-06
fix: remove subscription finalizers in mutation

a-cordier:


controllers/apim/subscription/subscription_controller.go
```diff
@@ -64,12 +64,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	events := event.NewRecorder(r.Recorder)
 
-	status := subscription.Status.DeepCopy()
+	dc := subscription.DeepCopy()
```
kamiiiel, line 67: Can you please make this consistent with other controllers. So we have the same pattern everywhere

PR #956, opened 2024-12-13
fix: solve the issue with load balancer type

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-862

### DIFF
--- a/api/model/api/v4/endpoints.go
+++ b/api/model/api/v4/endpoints.go
@@ -69,22 +69,15 @@ func NewHttpEndpoint(name string) *Endpoint {
 	}
 }
 
-// +kubebuilder:validation:Enum=ROUND_ROBIN;RANDOM;WEIGHTED_ROUND_ROBIN;WEIGHTED_RANDOM;
+// +kubebuilder:validation:Enum=round-robin;random;weighted-round-robin;weighted-random;
 type LoadBalancerType string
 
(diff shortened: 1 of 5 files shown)


api/model/api/v4/endpoints.go
```diff
@@ -69,22 +69,15 @@ func NewHttpEndpoint(name string) *Endpoint {
...
-	RoundRobin         LoadBalancerType = "ROUND_ROBIN"
-	Random             LoadBalancerType = "RANDOM"
-	WeightedRoundRobin LoadBalancerType = "WEIGHTED_ROUND_ROBIN"
-	WeightedRandom     LoadBalancerType = "WEIGHTED_RANDOM"
-)
-
 type LoadBalancer struct {
-	// +kubebuilder:default:=`ROUND_ROBIN`
+	// +kubebuilder:default:=`round_robin`
 	Type LoadBalancerType `json:"type"`
```
a-cordier, line 81: shouldn't this be `round-robin` given the issue description ?

a-cordier: also, I tested setting `random` with a management context, and the API still shows ROUND_ROBIN

a-cordier (changes requested):
Looks like the issue is not in the operator but at the export (or import ?) level because for the REST API the accepted value is indeed upper snake case

PR #961, opened 2024-12-17
fix: enhance management context admission error

a-cordier:
see https://gravitee.atlassian.net/browse/GKO-872


ctschacher (approved):
LGTM

PR #1056, opened 2025-02-13
feat: add gateway api to embedded crds

a-cordier:


hack/make/dev.mk
```diff
@@ -21,12 +21,16 @@ ifndef ignore-not-found
...
+	@kubectl apply -f helm/gko/crds/gravitee.io
+
+.PHONY: install-std-gateway-api
+install-gateway-api: ## Install gateway-api CRDs into the current cluster
+	@kubectl apply -f helm/gko/crds/kubernetes.io/gateway-api/standard
 
 .PHONY: uninstall
 uninstall: ## Uninstall CRDs from the current cluster
-	@kubectl delete -f helm/gko/crds
+	@kubectl delete -f helm/gko/crds/gravitee.io
```
benoitgravitee, line 34: Maybe not very useful, but uninstall would be nice to start from a clean slate if needed

PR #1082, opened 2025-03-06
fix: make endpoint qos optional for V4 APIs

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-1029


a-cordier (approved):

a-cordier:
Looks good to me, but I wonder if making AUTO the default in the schema would not be better in term of user documentation 🤔

PR #1096, opened 2025-03-18
fix: tidy up project (wip)

benoitgravitee:
- move package markers to dedicated files
- move tooling related files to tooling/
- move indexer to search package
- update linters
- add prettier check for scripts
- fix make helm-test


a-cordier:
could/should we use the hack package to put tooling / binary version management ? It looks like this what the kubernetes project does https://pkg.go.dev/k8s.io/kubernetes/hack

a-cordier (approved):

PR #1167, opened 2025-05-09
fix: do not call admission webhook update on CRD deletion

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-1208

### DIFF
--- a/internal/admission/group/crtl.go
+++ b/internal/admission/group/crtl.go
@@ -16,6 +16,7 @@ package group
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


internal/admission/group/crtl.go
```diff
@@ -56,5 +57,13 @@ func (a AdmissionCtrl) ValidateDelete(
 func (a AdmissionCtrl) ValidateUpdate(
 	ctx context.Context, oldObj runtime.Object, newObj runtime.Object,
 ) (admission.Warnings, error) {
+	group, ok := newObj.(*v1alpha1.Group)
+	if !ok {
+		return admission.Warnings{}, fmt.Errorf("wrong object")
```
benoitgravitee, line 62: maybe something more explicit, like `"wrong object was expecting v1alpha Group but got %s"`
something like that

internal/admission/resource/ctrl.go
```diff
@@ -64,5 +65,13 @@ func (a AdmissionCtrl) ValidateUpdate(
 	oldObj runtime.Object,
 	newObj runtime.Object,
 ) (admission.Warnings, error) {
+	res, ok := newObj.(*v1alpha1.ApiResource)
+	if !ok {
+		return admission.Warnings{}, fmt.Errorf("wrong object")
```
benoitgravitee, line 70: same as for groups

internal/admission/subscription/ctrl.go
```diff
@@ -56,5 +57,13 @@ func (a AdmissionCtrl) ValidateDelete(
 func (a AdmissionCtrl) ValidateUpdate(
 	ctx context.Context, oldObj runtime.Object, newObj runtime.Object,
 ) (admission.Warnings, error) {
+	subscription, ok := newObj.(*v1alpha1.Subscription)
+	if !ok {
+		return admission.Warnings{}, fmt.Errorf("wrong object")
```
benoitgravitee, line 62: same as for groups

benoitgravitee (approved):
Added a few comment to make errors more user friedly if that make sense

PR #1216, opened 2025-06-13
Use v1 API for Applications

Sander0542:
Applications are only accessible by the v1 API. The v2 API does not contain any application endpoints.

Reference: https://documentation.gravitee.io/apim/management-api/management-api-reference


Sander0542:
The admission webhook controller is giving me an 404 error on the application dryrun endpoint. What could this be related to then?

PR #1264, opened 2025-08-01
feat: add status condition to API CRDs

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-1248


benoitgravitee (changes requested):
Team decision during QBR easy to miss it, but here is the link: 
https://gravitee.atlassian.net/browse/GKO-1247?focusedCommentId=46737

a-cordier (approved):

PR #1291, opened 2025-08-18
test: adds e2e test to remove member from API

ctschacher:
This PR adds a new end-to-end test to verify correct behavior when removing a member from an API definition.

see: https://gravitee.atlassian.net/browse/GKO-1404

### DIFF
--- a/test/e2e/chainsaw/commands/deleteServiceAccount.mjs
+++ b/test/e2e/chainsaw/commands/deleteServiceAccount.mjs
@@ -1,0 +1,52 @@
+#!/usr/bin/env zx
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


test/e2e/chainsaw/commands/deleteServiceAccount.mjs
```diff
@@ -0,0 +1,52 @@
...
+import { mapiClient } from '../lib/mapiClient.mjs';
+
+const { name } = argv;
+if (!name) {
+  console.error('Usage: deleteServiceAccount.mjs --name <SERVICE_ACCOUNT_NAME>');
+  process.exit(1);
+}
+
+try {
+  const searchPath = `/management/organizations/DEFAULT/search/users?q=${encodeURIComponent(name)}`;
```
carlos-andres-osorio, line 27: Since `DEFAULT` is hard-coded, do we test that users can use other organizations in unit/integration tests?

Or will using different Organization IDs be supported in the future?

carlos-andres-osorio (approved):
Just a minor question, which is not a required change for this PR.

Looks good to me!

test/e2e/chainsaw/commands/deleteServiceAccount.mjs
ctschacher: Good question, thank you! I'll check how that is covered in other tests and also will consider to make the organisation id changeable.

PR #1319, opened 2025-09-02
feat: allow templating to be disabled

benoitgravitee:
## Issue

[GKO-1431](https://gravitee.atlassian.net/browse/GKO-1433)


benoitgravitee:
Conditions might not the weapon of choice for this. That means you add conditions to False on something that you disabled. I'll add a log instead if I can.

PR #1381, opened 2025-09-23
fix: make API type immutable

kamiiiel:
see: https://gravitee.atlassian.net/browse/GKO-1528


a-cordier:
Let's merge that one but I think we should consider moving this kind of validation to [CEL](https://kubernetes.io/docs/reference/using-api/cel/)

cc @kamiiiel @carlos-andres-osorio

PR #1396, opened 2025-09-25
test: introducing chainsaw scenarios

ctschacher:
This PR introduces Chainsaw Scenarios, a way to consolidate duplicated logic for v2 and v4 APIs into parameterized tests. It removes a lot of redundant files, simplifies resource naming, and improves maintainability by reducing duplication and making tests more generic. No tests were added, just refactoring to apply Scenarios in tests.

These changes will make it easier to add new API versions or test scenarios in the future and improve the maintainability of the test suite.

See: https://gravitee.atlassian.net/browse/GKO-1536


ctschacher:
@carlos-andres-osorio Thanks for your feedback! And I don't think that failure is linked to the changes, I checked again locally and it's perfectly fine. But we also had integration tests failing at the same time, not sure if there was something else going wrong. I'll re-start the test and see.

PR #1474, opened 2025-12-18
GKO to use Automation API

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-1418
https://gravitee.atlassian.net/browse/GKO-1968
https://gravitee.atlassian.net/browse/GKO-2558

## Description

* Controllers to use HRID (namespace-name) instead of generated IDs
* Use legacy* query params when former ID is present and keep on using it
* Handle general conditions and parent pages HRID properly
* Update all tests (integration/e2e) to use Automation API as much as possible
* Added a `failure.md` to list all test that cannot pass yet because of Automation API not managing Groups and Notification (4.12 roadmap)


a-cordier:
see https://gravitee.atlassian.net/browse/GKO-1968

a-cordier (approved):

PR #1633, opened 2026-04-09
fix: avoid subscription count increase on update

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2804

## Description

* Check if the subscription is created => increase subscription counter
* Add tests to make sure they are not updated when updating the subscription

No need to fix master, it is included in another PR.

### DIFF
--- a/api/v1alpha1/subscription_types.go
+++ b/api/v1alpha1/subscription_types.go
@@ -171,6 +171,14 @@ func (s *Subscription) SetConditions(conditions []metav1.Condition) {
 	s.Status.Conditions = conditions
 }
 
+func (s *Subscription) HasNoStatus() bool {
+	return s.Status.ID == "" &&
+		s.Status.StartedAt == "" &&
+		s.Status.EndingAt == "" &&
+		len(s.Status.Conditions) == 0 &&
+		s.Status.ProcessingStatus == ""
+}
+
 func init() {
 	SchemeBuilder.Register(&Subscription{}, &SubscriptionList{})
 }
--- a/test/integration/subscription/subscribe_withContext_updateEndDate_v4_test.go
+++ b/test/integration/subscription/subscribe_withContext_updateEndDate_v4_test.go
@@ -1,0 +1,122 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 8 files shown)


test/integration/subscription/subscribe_withContext_updateEndDate_v4_test.go
```diff
@@ -0,0 +1,96 @@
...
+			err := manager.GetLatest(ctx, fixtures.APIv4)
+			if err != nil {
+				return err
+			}
+			return assert.Equals(
+				"subscriptionCount",
+				1,
+				fixtures.APIv4.Status.SubscriptionCount,
+			)
+		}, constants.ConsistentTimeout, interval).ShouldNot(Succeed())
```
a-cordier: not sure, but is it really should not succeed here ? I have the feeling that the expected subscription count here is indeed 1 🤔

api/v1alpha1/subscription_types.go
```diff
@@ -171,6 +171,14 @@ func (s *Subscription) SetConditions(conditions []metav1.Condition) {
 	s.Status.Conditions = conditions
 }
 
+func (s *Subscription) HasNoStatus() bool {
+	return s.Status.ID == "" &&
+		s.Status.StartedAt == "" &&
+		s.Status.EndingAt == "" &&
+		len(s.Status.Conditions) == 0 &&
+		s.Status.ProcessingStatus == ""
```
a-cordier, line 179: Would an empty ID string be enough for this check ?

a-cordier:
I would suggest to rephrase the commit when everything all the comments are cleared

a-cordier:
ha, for the checks sorry, the job names have changed following a CI refactor. should be just fine on next push

api/v1alpha1/subscription_types.go
benoitgravitee: You are right. For now it is, but when Automation API kick in, it won't, I was sparing myself some time here :-)

test/integration/subscription/subscribe_withContext_updateEndDate_v4_test.go
benoitgravitee: Yes it is... I don't know why I did that... it makes question it all ... oh my. Good catch.

benoitgravitee: Subscription count test always failing to compare `int32(1)` and `uint(1)`.
I fixed the assertion and it is all OK.

a-cordier (approved):

PR #1654, opened 2026-04-24
feat: add support for customAPiKey subscription

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2548


kamiiiel:
All comments from Gemini are false positive, so I just ignored them

a-cordier (approved):

PR #1668, opened 2026-05-05
feat: use new GroupResource in automation API

kamiiiel:
see https://gravitee.atlassian.net/browse/GKO-2557

### DIFF
--- a/api/model/group/group.go
+++ b/api/model/group/group.go
@@ -32,6 +32,8 @@ const (
 type Type struct {
 	// +kubebuilder:validation:Optional
 	ID string `json:"id,omitempty"`
+	// +kubebuilder:skipversion
+	HRID string `json:"hrid,omitempty"`
 	// +kubebuilder:validation:Required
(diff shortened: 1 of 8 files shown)


api/model/group/group.go
```diff
@@ -32,6 +32,8 @@ const (
 type Type struct {
 	// +kubebuilder:validation:Optional
 	ID string `json:"id,omitempty"`
+	// +kubebuilder:validation:Optional
```
benoitgravitee: Better if not part of the CRD, so it does not appear in docs etc...
```suggestion
	// +kubebuilder:skipversion
```

benoitgravitee (changes requested):

api/model/group/group.go
kamiiiel: I updated the code but I didn't get what this marker suppose to do in here? This marker is for type ...

benoitgravitee (approved):

PR #1695, opened 2026-06-11
test(platform-test): provision the e2e cluster via gck

ctschacher:
## Summary

- Replace the run-kind.mjs flow for the **e2e (platform-test) job only** with [gck](https://gravitee-io-labs.github.io/gck/) **v1.0.2**: the job composes the `gravitee-io/oss/apim/mongodb` registry context. The APIM image source/tag (azurecr master nightlies) are recipe vars passed via `--set imagePrefix=graviteeio.azurecr.io --set imageTag=master-latest` in the new `make start-e2e-cluster` target; ES and portal are disabled via the recipe's flags. The integration-test job and local dev keep `make start-cluster`.
- `test/platform-test/gck.yaml` is intentionally minimal and only carries what recipe vars cannot express: the `/tmp/coverage` Kind mount backing the operator coverage PV, the chart version pin (4.12 prerelease-inclusive), and gateway kubernetes sync across all namespaces with the cluster RBAC the chart cannot render itself (its managed SA only gets a namespaced Role).
- APIM lives in the `gravitee` namespace, so 6 fixtures' ManagementContext `baseUrl` move from `apim-apim3-api.default.svc:83` to `apim-api.gravitee.svc:83`. The bearer-token templating fixture intentionally keeps an unreachable host (the admission webhook fails open when APIM is unreachable but would 401 the fixture's dummy token against a real endpoint).
- e2e README cluster-bootstrap section rewritten around gck; AGENTS.md references updated.

Two gck bugs found during this migration (invalid RSA key in the registry's tls-server secret, flag patches failing to resolve context template vars) were fixed upstream in [gravitee-io-labs/gck#2](https://github.com/gravitee-io-labs/gck/pull/2); this PR pins v1.0.2 and carries no workarounds.

Verified locally on a from-scratch `gck create`: full suite green (**345 passed, 5 pre-existing skips**), gateway/mAPI running the azurecr `master-latest-debian` nightlies, mongo unaffected by the broadcast `--set` (stays `mongo:7`), mTLS port 30084 handshaking with the registry cert.

See: https://gravitee.atlassian.net/browse/GKO-2948

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -452,10 +452,17 @@ jobs:
           registry: graviteeio.azurecr.io
           username: keeper://Q721P2LSOPJ9qiXLuf5AHQ/field/login
           password: keeper://Q721P2LSOPJ9qiXLuf5AHQ/field/password
+      # Must run after docker-login: the gck image preload pulls the APIM
+      # master nightlies from graviteeio.azurecr.io.
+      - run:
+          name: Install gck
+          command: |
+            go install github.com/gravitee-io-labs/gck@v1.0.2
+            echo 'export PATH="$HOME/go/bin:$PATH"' >> "$BASH_ENV"
       - run:
           name: Start cluster
           command: |
-            make start-cluster
+            make start-e2e-cluster
       - run:
           name: Prepare coverage storage
           command: |
--- a/hack/make/dev.mk
+++ b/hack/make/dev.mk
@@ -8,6 +8,11 @@ start-cluster: ## Init and start a local cluster
 start-cluster-ui: ## Init and start a local cluster
 	@APIM_UI=true npx zx ./hack/scripts/run-kind.mjs
 
+.PHONY: start-e2e-cluster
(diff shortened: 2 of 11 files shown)


.circleci/config.yml
```diff
@@ -452,10 +452,18 @@ jobs:
           registry: graviteeio.azurecr.io
           username: keeper://Q721P2LSOPJ9qiXLuf5AHQ/field/login
           password: keeper://Q721P2LSOPJ9qiXLuf5AHQ/field/password
+      # gck has no tagged releases yet — pin to a known-good commit and bump
+      # deliberately. Must run after docker-login: the gck image preload pulls
+      # the APIM master nightlies from graviteeio.azurecr.io.
+      - run:
+          name: Install gck
+          command: |
+            go install github.com/gravitee-io-labs/gck@0307d33a6e1a
```
a-cordier: why a commit sha ?

a-cordier (approved):

.circleci/config.yml
ctschacher: I updated that part, it will now take a release instead of a sha.

PR #1702, opened 2026-06-19
test(platform-test): foundation for e2e upgrade testing (version gating + before/after flow)

ctschacher:
Foundation for automated upgrade testing in the platform-test e2e suite: the
reusable parts (version gating, lane selection, before/after-upgrade
orchestration). The dedicated "do old resources survive the upgrade?" survival
check is a follow-up built on this (see below).

- **Lane selection now covers the full suite.** `--provision-with gko|terraform`
  runs that provisioner's whole lane (its `tests/<lane>/` folder plus the matching
  arm of every shared scenario), instead of only the migrated scenarios. The config
  ignores the other lane's folder and drops its arm with a case-sensitive grepInvert.
- **Version gating.** `--run-up-to-version <semver>` is enforced: an automatic
  fixture skips any test tagged `@since-<newer>` than the cap. The tag can sit on a
  single `describe()` to gate a whole file, and the matrix runner supports a
  per-provisioner `since`. Untagged tests are baseline and always run.
- **Before/after-upgrade CI flow.** A post-merge job stands up APIM on the old line
  with `gck create`, runs the suite "up to old", upgrades APIM in place with
  `gck patch`, then runs "up to new". Since gck v1.1.1, `gck patch` inherits the
  create-time context, so the upgrade is a short inline sequence of `gck`/`e2e`
  commands (no orchestration script). The APIM chart version is the apim recipe's
  native `helmVersion` var (default unchanged, so the normal start-e2e-cluster flow
  is untouched).
- **Labels.** The 4.12-only features are tagged `@since-4.12`: api-key plan
  subscriptions (both provisioners), the Terraform `apim_group` resource (Terraform
  arm only) and dictionaries.

What this is **not** yet: it runs the suite before and after the upgrade, but each
test self-cleans, so it does not prove that resources created on the old version
survive the upgrade and stay mutable. That survival check is the next piece, built
on this foundation via the provisioner `attach()` seam (follow-up).

Verified end-to-end on a local kind cluster: `gck create` 4.11 → run → `gck patch`
to 4.12 (Mongo preserved) → run, green. Everything lives under `test/platform-test/`
except the one CI job in `.circleci/config.yml`.

See: https://gravitee.atlassian.net/browse/GKO-1898

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -457,7 +457,7 @@ jobs:
       - run:
           name: Install gck
           command: |
-            go install github.com/gravitee-io-labs/gck@v1.0.2
+            go install github.com/gravitee-io-labs/gck@v1.1.1
             echo 'export PATH="$HOME/go/bin:$PATH"' >> "$BASH_ENV"
       - run:
           name: Start cluster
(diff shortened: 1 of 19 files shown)


.circleci/config.yml
```diff
@@ -557,6 +557,60 @@ jobs:
...
+      - kubernetes/install_kubectl:
+          kubectl_version: "<< pipeline.parameters.kubernetes-version >>"
+      - helm/install_helm_client:
+          version: "<< pipeline.parameters.helm-version >>"
+      - run:
+          name: Install kind
+          command: |
+            curl -Lo /tmp/kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+            chmod +x /tmp/kind
+            sudo mv /tmp/kind /usr/local/bin/
```
a-cordier, line 578: you don't have to install kind, it's bundled in gck :)

.circleci/config.yml
```diff
@@ -557,6 +557,60 @@ jobs:
...
+      docker_layer_caching: true
+    resource_class: large
+    steps:
+      - checkout
+      - go/install:
+          version: "<< pipeline.parameters.go-version >>"
+      - kubernetes/install_kubectl:
+          kubectl_version: "<< pipeline.parameters.kubernetes-version >>"
+      - helm/install_helm_client:
+          version: "<< pipeline.parameters.helm-version >>"
```
a-cordier, line 572: you don't need to install helm, it's bundled in gck :)

.circleci/config.yml
```diff
@@ -557,6 +557,60 @@ jobs:
...
+    machine:
+      image: ubuntu-2204:2024.01.2
+      docker_layer_caching: true
+    resource_class: large
+    steps:
+      - checkout
+      - go/install:
+          version: "<< pipeline.parameters.go-version >>"
+      - kubernetes/install_kubectl:
+          kubectl_version: "<< pipeline.parameters.kubernetes-version >>"
```
a-cordier, line 570: you don't need to install kubectl, it's bundled in gck :)

test/platform-test/scripts/e2e-upgrade.mjs
```diff
@@ -0,0 +1,142 @@
...
+const mongoAfter = capture("kubectl", ["get", "pod", "-n", "gravitee", "-l", "app=mongodb",
+  "-o", "jsonpath={.items[0].metadata.name}"]);
+if (mongoBefore && mongoAfter && mongoBefore !== mongoAfter) {
+  throw new Error(`MongoDB pod was recreated by the upgrade (${mongoBefore} -> ${mongoAfter}); the database must survive in place.`);
+}
+console.log(`MongoDB pod preserved across the upgrade: ${mongoAfter || "(not found)"}`);
+
+// 5. Regression run on the NEW version.
+runSuite(NEW_VERSION, `after upgrade, APIM ${NEW_VERSION}`);
+
```
a-cordier: There's a lot going on here WDYT ?

Why are we testing that mongo is not recreated for instance, did you have an issue with this ?

There is an operator context in gck ...

a-cordier (changes requested):

a-cordier:
It's cool 👍 That being said I really think we should get up to a point where we don't need a script and the ci just looks like (the flags are not the right ones but they exist)

```
gck create --from "..." --set apim.imageTag=4.11-latest --set helm.version=4.11.0
e2e --provision-with gko run-up-to 4.11.0
gck patch gravitee --set  apim.imageTag=4.12-latest --set helm.version=4.12.0
e2e --provision-with gko run-up-to 4.12.0
```

If there are blocker on gck just let me now

test/platform-test/scripts/e2e-upgrade.mjs
a-cordier: https://gravitee-io-labs.github.io/gck/registry/gravitee-io/oss/gko/

a-cordier (changes requested):

.circleci/config.yml
ctschacher: Ok, will drop that.

.circleci/config.yml
ctschacher: Ok, will drop that.

.circleci/config.yml
ctschacher: The test suite still needs kubectl on PATH for its assertions. Does gck put its bundled kubectl on PATH, or should I keep just that install?

test/platform-test/scripts/e2e-upgrade.mjs
ctschacher: Agreed, this is too heavy. Almost all of it is working around step 4: `gck patch` re-introduced Elasticsearch (so I added `--disable-es`), it resolved images from the wrong registry (so I added `--set imagePrefix` + a patch file), and I built/loaded/installed GKO by hand.

How I'd simplify it:
- Drop the Mongo check. No, I never hit a Mongo issue, it was just defensive and you're right that it tests gck rather than the operator
- Use the `gravitee-io/oss/gko` context with a `builds:` entry so gck builds and deploys the branch operator, instead of the manual docker-build / kind-load / helm-install
- Delete the script and inline `create -> e2e -> patch -> e2e` in the CI job

For that last step I need the correct per-component `--set` flags so `gck patch` only bumps APIM's chart version + image tag (Mongo untouched, azurecr registry, no ES wait). What are they? Happy to pair on the gck.yaml.

ctschacher:
Ok, yes I agree, that four-line shape must be the target and I'll try to simplify that.

The only real blocker is the exact `patch` flags. Your sketch is basically what I want; mine got a bit more complicated because `gck patch` re-introduced Elasticsearch (so I added `--disable-es`) and pulled images from the wrong registry (so I added `--set imagePrefix` plus a one-line patch file). 

What's the correct way to have `gck patch gravitee --set apim.imageTag=4.12-latest --set ...` bump only APIM's chart version + image tag, leave Mongo in place, and stay on azurecr? 

With that I'll drop the script and deploy the branch operator through the `gravitee-io/oss/gko` context.

(Specifics are on the e2e-upgrade.mjs thread.)

a-cordier (approved):

PR #1713, opened 2026-06-30
feat(GKO-2995): add drift detection framework

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2995

## Summary

- Add `internal/drift` package: struct-tag-driven comparison engine (`Detect`), pluggable equivalence registry (`empty-is-nil`, `trimmed`, `rfc3339`, `ignore`, `unstructured`), merge logic for admission (old/new CRD vs remote), and pseudo-YAML drift output.
- Add `internal/admission/drift` admission helper (`ValidateDrift`) to compare resolved CRD DTOs against remote APIM state and reject updates when external drift is detected.
- Wire operator startup (`drift.Init()`), env config (`ENABLE_DRIFT_DETECTION`, enabled by default), per-resource annotation (`gravitee.io/drift-detection`), and Helm values (`manager.driftDetection.enabled`).
- Add package documentation (`internal/drift/doc.go`) and unit test suite under `test/unit/drift/`.

## Test plan

- [x] `make unit` (drift suite: detect, equivalences, registry — 242 specs)
- [x] `go build .`
- [x] Helm unittest: `helm unittest helm/gko -f tests/manager/config_test.yaml`


internal/drift/types.go
```diff
@@ -0,0 +1,157 @@
...
+	if len(or.Children) > 0 {
+		merger.Children = make([]*Result, len(or.Children))
+		for i, orChild := range or.Children {
+			// create a new child
+			merger.Children[i] = &Result{}
+			// evaluate the child and merge it
+			merge(*orChild, *nr.Children[i], merger.Children[i])
+		}
+	}
+}
```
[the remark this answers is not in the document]
benoitgravitee: Won't do, useless in the context of use, not lib just a catered for GKO

internal/drift/detect.go
```diff
@@ -0,0 +1,455 @@
...
+			// handle the field as a simple value
+			equivalenceFunc := equivalenceRegistry.Get(funcName, fieldType.Kind())
+			equivalent := equivalenceFunc(crdPair.Interface, remotePair.Interface)
+			this.AppendChild(&Result{
+				Property:    property,
+				Equivalence: equivalent,
+				CRDValue:    crdPair.Interface,
+				RemoteValue: remotePair.Interface,
+			}, ordered)
+		}
```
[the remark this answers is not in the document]
benoitgravitee: Not happening for GKO, we don't and will never have "any" for a CRD field type, it won't work

internal/drift/types.go
```diff
@@ -0,0 +1,157 @@
...
+	merger.CRDValue = result.CRDValue
+	merger.RemoteValue = result.RemoteValue
+	// merge children
+	if len(or.Children) > 0 {
+		merger.Children = make([]*Result, len(or.Children))
+		for i, orChild := range or.Children {
+			// create a new child
+			merger.Children[i] = &Result{}
+			// evaluate the child and merge it
+			merge(*orChild, *nr.Children[i], merger.Children[i])
```
carlos-andres-osorio: `merge()` pairs `or.Children`/`nr.Children` positionally (`nr.Children[i]`), assuming the O/R and N/R `Result` trees are shape-identical. `Detect` tree shape depends on CRD content (slice length via `detectItems`/`detectAsymmetrical`, map key sets via `detectIndexedMapItems`), and `ValidateDriftWithContext` runs `Detect` on two independently user-supplied CRD revisions (old vs new) against the same remote snapshot, so the trees diverge whenever the update changes a slice length or map key set.

## Evidence

Real-CRD repro using `v4.Flow` (`spec.flows[].request`), parsed from three YAML revisions into the real Go type:

`old-flow.yaml` (4 request steps):
```yaml
name: default
enabled: true
request:
  - policy: rate-limit
    enabled: true
  - policy: cors
    enabled: true
  - policy: key-less
    enabled: true
  - policy: transform-headers
    enabled: true
```

`new-flow.yaml` / `remote-flow.yaml` (1 request step, identical to each other):
```yaml
name: default
enabled: true
request:
  - policy: rate-limit
    enabled: true
```

```go
orResult := drift.Detect(oldFlow, remoteFlow)   // 15 children
nrResult := drift.Detect(newFlow, remoteFlow)   // 12 children
drift.Merge(orResult, nrResult)                 // panics
```

```
panic: runtime error: index out of range [12] with length 12
	internal/drift.merge(...) types.go:135
	internal/drift.Merge(...)
```

A routine "the applied CRD removed some flow steps" update panics the admission webhook.

internal/drift/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// # Extending
+//
+// Register additional equivalence functions with [RegisterEquivalenceFunc] and call
+// them from [Init]. Annotate api/model fields with the matching drift tag name.
+//
+// Reference fixtures and behaviour tables live in test/unit/drift/.
+//
+// # Enabling drift detection
+//
+// Globally via the ENABLE_DRIFT_DETECTION environment variable (enabled by default).
```
carlos-andres-osorio, line 93: This section documents the global `ENABLE_DRIFT_DETECTION` env var and the per-resource `gravitee.io/drift-detection` annotation as independent, symmetric switches. In reality, `drift.Init()` (which registers every named `EquivalenceFunc`) is only called in `main.go:182` when the *global* flag is true, while `isDriftDetectionEnabled` (`internal/admission/drift/validate.go:109`) still honors the annotation when the global flag is off. In that combination the registry is empty and `registry.Get` calls `log.Panicf` for any field carrying a non-empty `drift:"..."` tag.

## Evidence

Doc text (lines 93-96):
```
// Globally via the ENABLE_DRIFT_DETECTION environment variable (enabled by default).
// Per resource via the gravitee.io/drift-detection annotation (true/false).
```
No mention that the annotation only works when `drift.Init()` has already run, which only happens when the global flag was true at startup (`main.go:182`). Confirmed via a real `ApiV4Definition` with `gravitee.io/drift-detection: "true"` and `env.Config.DriftDetection = false`, run through the real `ValidateDriftWithContext` with a stand-in tagged DTO:

```
panic: drift function 'empty-is-nil' not found for kind 'string'
```

internal/drift/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// [Equivalence] with a status, an optional [Equivalence.Skip] flag, and an optional
+// [Equivalence.PostFunc] hook.
+//
+// Registered names (see [Init]):
+//
+//   - empty-is-nil (string): nil or "" are equivalent; otherwise deep-equal.
+//   - empty-is-nil (bool): nil or false are equivalent; otherwise deep-equal.
+//   - empty-is-nil (int): nil or 0 are equivalent; otherwise deep-equal.
+//   - empty-is-nil (uint): nil or 0 are equivalent; otherwise deep-equal.
+//   - empty-is-nil (slice, map): same length is equivalent at the container level;
```
carlos-andres-osorio: The `empty-is-nil (slice, map)` bullet states "item-level comparison continues unless lengths differ ([CannotCompare])", implying item comparison is skipped when lengths differ. `EmptyIsNilLen` never sets `Equivalence.Skip` (whether lengths match or not), so `detectStruct` always recurses into slice/map items regardless of length equality — comparison never actually stops. The doc describes the opposite of the real behavior for the "differ" case.

## Evidence

Doc text (line 62-63):
```
//   - empty-is-nil (slice, map): same length is equivalent at the container level;
//     item-level comparison continues unless lengths differ ([CannotCompare]).
```
`EmptyIsNilLen` (`internal/drift/equivalences.go:131`) never sets `Skip: true` in any branch — it returns `Equivalence{Equivalent: CannotCompare}` unconditionally at the end, which `detectStruct`'s `if !equivalent.Skip` check always satisfies. Item comparison runs every time, contradicting the "unless lengths differ" phrasing.

main.go
```diff
@@ -178,17 +179,15 @@ func main() {
 		}
 	}
 
-	//+kubebuilder:scaffold:builder
-
-	if healthCheckErr := mgr.AddHealthzCheck("healthz", healthz.Ping); healthCheckErr != nil {
-		log.Global.Error(healthCheckErr, "Unable to set up health check")
-		os.Exit(1)
+	if env.Config.DriftDetection {
+		drift.Init()
```
carlos-andres-osorio: `drift.Init()` is only called when `env.Config.DriftDetection` is true, but `isDriftDetectionEnabled` lets the `gravitee.io/drift-detection: "true"` annotation re-enable drift detection per-resource when the global flag is off — a contract documented in `internal/drift/doc.go:96` and `helm/gko/README.md:112`. In that state the equivalence registry is empty and `registry.Get` panics via `log.Panicf` for any field carrying a `drift:"..."` tag. Latent today (no `drift:"..."` tags exist yet on `api/model` DTOs) but detonates the moment the first tagged DTO lands.

## Evidence

Real CRD, annotation-only enablement with the global flag off:

```yaml
kind: ApiV4Definition
metadata:
  name: my-api
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  name: my-api
  apiVersion: "1.0"
  type: PROXY
```

```go
env.Config.DriftDetection = false // simulates ENABLE_DRIFT_DETECTION=false at startup -> drift.Init() never runs
// crd parsed from the YAML above, run through the real admissiondrift.ValidateDriftWithContext(...)
// with a stand-in drift:"empty-is-nil"-tagged DTO field, since no real DTO field carries a drift tag yet
```

```
panic: drift function 'empty-is-nil' not found for kind 'string'
```

internal/drift/registry.go
```diff
@@ -0,0 +1,113 @@
...
+	} else if name != "" {
+		log.Panicf("drift function '%s' not found for kind '%s'", name, k)
+	}
+	if k == reflect.Slice {
+		return defaultSliceEquivalence
+	}
+	if k == reflect.Struct {
+		return defaultStructEquivalence
+	}
+	return FromDeepEqual
```
carlos-andres-osorio, line 56: Untagged map fields fall through `Get("", reflect.Map)` to `FromDeepEqual`. Combined with `detectStruct`'s fall-through (only the struct branch has an `else`; the map-items branch and the trailing "simple value" branch both run for the same field), every untagged map field gets a second whole-map child compared with `reflect.DeepEqual`. Consequences: a nil CRD map vs an empty remote map is `Inequivalent` in both O/R and N/R trees, a permanent false-positive drift the user cannot fix; and any genuine map difference is double-reported.

## Evidence

Real CRD field: `dictionary.ManualSpec.Properties map[string]string` (a real `Dictionary` CRD's `spec.manual.properties`, untagged).

**Consequence 1 — nil vs empty map:**
```yaml
# crd
kind: Dictionary
metadata:
  name: my-dictionary
spec:
  name: my-dictionary
  deployed: true
  type: MANUAL
  manual: {}          # properties absent -> nil map
```
```yaml
# remote (equivalent state, explicitly empty)
manual:
  properties: {}       # explicitly empty map
```
```
result := drift.Detect(crd.Spec.Manual, remote.Spec.Manual)
DriftDetected: true
result tree: "properties: map[] != map[]"
```
A nil vs empty map is reported as drift with no way to reconcile it — the CRD YAML has no way to express "an explicitly empty map" vs "absent" once serialized, so the user cannot fix this by editing their CRD.

**Consequence 2 — genuine one-key change, double-reported:**
```yaml
# crd
manual:
  properties:
    rateLimit: "100"
```
```yaml
# remote
manual:
  properties:
    rateLimit: "200"
```
Produced **2 children** for the single changed field instead of 1 — the correct per-key diff plus a spurious duplicate whole-map comparison.

internal/drift/detect.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+}
+
+func handleStructField(property string, funcName string, field reflect.StructField, crd any, remote any, this *Result) {
+	if isEmbeddedStruct(field) {
+		// no child creation we want fields to be flattened as it is embedded
+		detectStruct(crd, remote, this, false)
+	} else {
+		equivalenceFunc := equivalenceRegistry.Get(funcName, reflect.Struct)
+		this.Equivalence = equivalenceFunc(crd, remote)
```
carlos-andres-osorio: `handleStructField` writes `this.Equivalence = equivalenceFunc(crd, remote)` onto the *parent* result, which is overwritten by every subsequent struct field of the same parent. For untagged pointer-to-struct fields where `defaultStructEquivalence` returns `Inequivalent, Skip=true` (exactly one side nil), no child is appended, and either the drift is silently lost (overwritten by a later field) or, if it's the last field, the parent ends up `Inequivalent` with no property/values recorded. Also breaks `Merge` shape symmetry, feeding the `types.go:135` finding.

## Evidence

Real CRD field: `v4.EndpointGroupServices` (`spec.endpointGroups[].services`, has two untagged pointer-to-struct fields, `Discovery` and `HealthCheck`) — remote has both configured, the CRD revision drops `discovery`:

```yaml
# remote — both services configured
kind: ApiV4Definition
metadata:
  name: my-api
spec:
  name: my-api
  apiVersion: "1.0"
  type: PROXY
  endpointGroups:
    - name: default-group
      services:
        discovery:
          enabled: true
          overrideConfiguration: false
        healthCheck:
          enabled: true
          overrideConfiguration: false
```
```yaml
# crd — discovery dropped on re-apply
kind: ApiV4Definition
metadata:
  name: my-api
spec:
  name: my-api
  apiVersion: "1.0"
  type: PROXY
  endpointGroups:
    - name: default-group
      services:
        healthCheck:
          enabled: true
          overrideConfiguration: false
```

```go
result := drift.Detect(crd.Spec.EndpointGroups[0].Services, remote.Spec.EndpointGroups[0].Services)
```

The `Discovery` XOR-nil field overwrites the parent's `Equivalence` to `Inequivalent, Skip=true` and appends no child — the drift (discovery configured on the remote gateway but dropped from the CRD) is present in `DriftDetected()` but invisible in `Result.String()`: nothing about `discovery` appears in the rendered tree at all.

internal/drift/detect.go
```diff
@@ -0,0 +1,454 @@
...
+				Property:    key,
+				Equivalence: equivalent,
+				CRDValue:    crd.Interface,
+				RemoteValue: remote.Interface,
+			}, true)
+		}
+	}
+}
+
+func detectAny(key string, funcName string, crd valuePair, remote valuePair, parent *Result) bool {
```
carlos-andres-osorio, line 370: `detectAny` guards only against `reflect.Invalid` counterparts. When two `any`-typed values have different non-invalid kinds (reachable via `unstructured.Unstructured` configs whenever the remote value's type differs from the CRD's, e.g. drift by type change), it recurses with mismatched values — `detectStruct(string, struct)` panics in `t.NumField()`, `detectMapItems` panics in `MapKeys()` on a non-map value. The webhook then fails the request with a panic instead of reporting the drift.

## Evidence

Real CRD, `GenericStringMap`-backed `spec.flows[].request[].configuration` where the same key holds a different JSON kind on each side (object vs string):

```yaml
# old / crd
kind: ApiV4Definition
metadata:
  name: rate-limited-api
spec:
  name: rate-limited-api
  apiVersion: "1.0"
  type: PROXY
  flows:
    - name: default
      enabled: true
      request:
        - policy: rate-limit
          enabled: true
          configuration:
            timeout:
              value: 5000
              unit: ms
```
```yaml
# new / remote
kind: ApiV4Definition
metadata:
  name: rate-limited-api
spec:
  name: rate-limited-api
  apiVersion: "1.0"
  type: PROXY
  flows:
    - name: default
      enabled: true
      request:
        - policy: rate-limit
          enabled: true
          configuration:
            timeout: "5s"
```

`configuration.timeout` is an object on one side (`{value, unit}`) and a bare string (`"5s"`) on the other — exactly the "drift by type change" scenario the finding describes. `detectAny` recurses into `detectStruct`/`detectMapItems` with the mismatched-kind pair instead of reporting a leaf `Inequivalent`, panicking inside the reflect call rather than surfacing a drift result.

internal/drift/types.go
```diff
@@ -0,0 +1,157 @@
...
+	// Case 2
+	if nr.Equivalent == Equivalent {
+		return nr
+	}
+	// Case 4
+	if or.Equivalent == Equivalent {
+		return or
+	}
+	// should not happen but just in case
+	return Result{}
```
carlos-andres-osorio, line 226: The `evaluateDrift` fallback returns an empty `Result{}` when one tree reports `CannotCompare` and the other `Inequivalent` — reachable, not a "should not happen" case. When it fires, `merge` copies the zero `Result` into the merged node, erasing `Property`, `Index`, `CRDValue`, and `RemoteValue`, so the rendered drift tree shows a nameless node with no values, and a detected difference is silently discarded.

## Evidence

Real CRD field: `v4.EndpointGroupServices`, three revisions — old CRD and remote both have `healthCheck` configured, new CRD drops it:

```yaml
# old-crd.yaml / remote.yaml (identical)
kind: ApiV4Definition
metadata:
  name: repro-merge-fallback
spec:
  hrid: repro-merge-fallback
  type: PROXY
  listeners: []
  endpointGroups:
    - name: default-group
      services:
        discovery:
          enabled: false
        healthCheck:
          enabled: true
```
```yaml
# new-crd.yaml
kind: ApiV4Definition
metadata:
  name: repro-merge-fallback
spec:
  hrid: repro-merge-fallback
  type: PROXY
  listeners: []
  endpointGroups:
    - name: default-group
      services:
        discovery:
          enabled: false
```

```go
orResult := drift.Detect(oldServices, remoteServices)  // root Equivalent = CannotCompare (0)
nrResult := drift.Detect(newServices, remoteServices)  // root Equivalent = Inequivalent (2)
drift.Merge(orResult, nrResult)
```

Confirmed the mixed-status precondition is real and reachable via this CRD edit (`O/R status: 0`, `N/R status: 2`). In practice this branch's own symptom currently can't be observed standalone: the same field flip that trips this mismatch also makes `or.Children`/`nr.Children` different lengths at that same node, so `Merge` panics (`types.go:135`, filed separately) in the same call before this node's corrupted-but-non-panicking output could ever be returned or rendered. Still a real, independently-fixable bug — just currently masked by the other one.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,121 @@
...
+	if !isDriftDetectionEnabled(newCRD) {
+		return errs
+	}
+
+	oldCopy := oldCRD.DeepCopyObject()
+	newCopy := newCRD.DeepCopyObject()
+
+	// We need template to be compiled
+	if err := admission.CompileAndValidateTemplate(ctx, oldCopy); err != nil {
+		errs.AddWarningf("could not compile templates of existing CRD, drift might be detected: %s", err.Error())
```
carlos-andres-osorio: Every other admission validator in this codebase (`portallisting`, `policygroups`, `mctx`, `docs`, `ctxref`, `subscription`, `api/base`, `resource`, `application`) calls `errs.Add(admission.CompileAndValidateTemplate(...))`, which preserves whatever severity `CompileAndValidateTemplate` assigns — and it always returns `errors.NewSeveref(...)` on failure (`internal/admission/template.go:32`), i.e. blocking. This is the only caller that discards the actual error and instead calls `errs.AddWarningf(...)`, downgrading a template-compile failure to a non-blocking warning, then continues on to drift detection using the very copy whose template failed to compile.

## Evidence

Real CRD with a malformed EL template expression:

```yaml
kind: ApiV4Definition
metadata:
  name: repro-severity-downgrade
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  hrid: repro-severity-downgrade
  type: PROXY
  description: "[[ .BadTemplateSyntax"
  listeners: []
```

```go
env.Config.EnableTemplating = true
// crd parsed from the YAML above, run through the real ValidateDriftWithContext
```

```
Direct admission.CompileAndValidateTemplate on the same broken CRD: template: gko:1: unclosed action
ValidateDriftWithContext result: Severe=0 Warning=2 IsSevere=false
Would this BLOCK admission? <nil> (nil means allowed through)
```

The template fails to compile on both the old and new CRD copies (2 warnings), yet `IsSevere()` is false and `Map()` returns a nil error — admission succeeds anyway, unlike every other resource type in this operator faced with the same malformed template.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	if err := admission.CompileAndValidateTemplate(ctx, newCopy); err != nil {
+		errs.AddWarningf("could not compile templates of updated CRD, drift might be detected: %s", err.Error())
+	}
+
+	// We need to resolve all references to compare the content that sent to the remote API
+	resolveRefs(ctx, oldCopy, errs)
```
carlos-andres-osorio: After `resolveRefs(ctx, oldCopy, errs)` / `resolveRefs(ctx, newCopy, errs)` run (which can append `Severe` entries per the `RefResolver` contract), execution is not short-circuited: the code falls straight through to `resolveContext`, `getRemoteObject`, `mapDTO`, and `drift.Detect` regardless of whether ref resolution failed — inconsistent with the very next error check (`resolveContext`), which returns immediately on failure.

## Evidence

Real CRD:

```yaml
kind: ApiV4Definition
metadata:
  name: repro-refs-shortcircuit
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  hrid: repro-refs-shortcircuit
  type: PROXY
  listeners: []
```

```go
// a realistic RefResolver: fails to resolve a referenced SharedPolicyGroup
brokenRefResolver := func(_ context.Context, _ runtime.Object, errs *errors.AdmissionErrors) {
    errs.AddSeveref("SharedPolicyGroup %q not found", "missing-spg")
}
// getRemoteObject is a spy that records whether it was called
admissiondrift.ValidateDriftWithContext(ctx, &crd, &crd, stubContext, brokenRefResolver, spyRemoteObject, mapDTO)
```
```
resolveRefs recorded Severe errors: 2 (ran for both old and new copy)
getRemoteObject was called AFTER resolveRefs failed: true
```

A wasted remote APIM call runs against an object with a known-broken reference, and drift comparison can add further errors on top of the actual ref-resolution failure.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,121 @@
...
+	resolveRefs(ctx, newCopy, errs)
+
+	apimClient, err := resolveContext(ctx)
+	if err != nil {
+		errs.AddSeveref("could not resolve context for CRD: %s", err.Error())
+		return errs
+	}
+
+	remoteObject, ok := getRemoteObject(apimClient, newCopy, errs)
```
carlos-andres-osorio: Minor: `remoteObject, ok := getRemoteObject(apimClient, newCopy, errs)` — when `ok` is `false`, the function trusts that the caller-supplied `RemoteObjectGetter` implementation already recorded an entry in `errs` explaining why, but nothing enforces or verifies that contract.

## Evidence

Real CRD (same fixture as the `validate.go:82` finding):

```yaml
kind: ApiV4Definition
metadata:
  name: repro-refs-shortcircuit
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  hrid: repro-refs-shortcircuit
  type: PROXY
  listeners: []
```

```go
silentNotFoundGetter := func(_ *apim.APIM, _ runtime.Object, _ *errors.AdmissionErrors) (any, bool) {
    return nil, false // realistic mistake: forgets to record why before returning
}
```
```
Severe=0 Warning=0
Admission blocked? <nil>
```

If a future `RemoteObjectGetter` implementation ever returns `false` without populating `errs`, admission succeeds silently with zero indication drift detection was skipped. Worth a defensive assert/log here so a future implementation bug fails loudly instead of silently — not blocking, just a robustness nit.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,121 @@
...
+		return apim.FromContextRef(ctx, newCRD.ContextRef(), newCRD.GetNamespace())
+	}, resolveRefs, getRemoteObject, dtoMapper)
+}
+
+func ValidateDriftWithContext(
+	ctx context.Context,
+	oldCRD runtime.Object,
+	newCRD runtime.Object,
+	resolveContext ContextResolver,
+	resolveRefs RefResolver,
```
carlos-andres-osorio, line 64: `ValidateDriftWithContext` (and its wrapper `ValidateDrift`) has no test coverage anywhere in this diff — no test file exists for `internal/admission/drift`. This is the actual wiring point between the drift engine and the admission webhook, so a bug here is the one place regressions would reach production undetected by the otherwise-thorough `internal/drift` unit tests.

Distinct branches with no regression protection: the `!isDriftDetectionEnabled` short-circuit, the two template-compile-failure paths that add warnings but continue, the `resolveContext` error path, the `getRemoteObject` "not found" path, the `drift.Merge(...).DriftDetected()` severe-error path, and the implicit "no drift" empty-`errs` path.

Confirmed no test directory exists at `test/unit/admission/drift/` at all.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,121 @@
...
+	newVsRemoteResult := drift.Detect(newDTO, remoteObject)
+
+	if result := drift.Merge(oldVsRemoteResult, newVsRemoteResult); result.DriftDetected() {
+		errs.AddSeveref("\ndrift detected:\n%s", result.String())
+	}
+
+	return errs
+}
+
+func isDriftDetectionEnabled(newCRD runtime.Object) bool {
```
carlos-andres-osorio: `isDriftDetectionEnabled` has three branches (annotation `"true"`, annotation `"false"`, no annotation falling back to `env.Config.DriftDetection`) plus a `log.Panicf` guard when `newCRD` doesn't implement `core.Object`. None of these branches are exercised by any test in the diff.

## Evidence

Confirmed no test file exists for this package. The `main.go:183` finding's evidence in this review incidentally exercised one of these branches as a throwaway repro (not permanent coverage) — a real CRD with the annotation set and the global flag off:

```yaml
kind: ApiV4Definition
metadata:
  name: my-api
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  name: my-api
  apiVersion: "1.0"
  type: PROXY
```

The `"false"` annotation branch, the no-annotation-fallback branch, and the non-`core.Object` panic remain entirely untested.

helm/gko/templates/manager/config.yaml
```diff
@@ -124,3 +124,8 @@ data:
   {{- else }}
   ENABLE_TEMPLATING: "false"
   {{- end }}
+  {{- if .Values.manager.driftDetection.enabled }}
```
carlos-andres-osorio, line 127: No test for the `driftDetection.enabled=false` branch of this template, which must render `ENABLE_DRIFT_DETECTION: "false"` explicitly (unlike most flags in this template, the else-branch emits an explicit `"false"` rather than omitting the key). `config_test.yaml` only asserts the default (`"true"`) — the analogous metrics enabled/disabled pair in the same test suite does have both cases.

## Evidence

```
config.yaml:127   {{- if .Values.manager.driftDetection.enabled }}
config.yaml:128     ENABLE_DRIFT_DETECTION: "true"
config.yaml:130     ENABLE_DRIFT_DETECTION: "false"
```

Only test case in `config_test.yaml`: `it: Templating and drift detection enabled by default`, asserting `data.ENABLE_DRIFT_DETECTION == "true"`. No case anywhere in the suite sets:

```yaml
# helm values fragment that would exercise the untested else-branch
manager:
  driftDetection:
    enabled: false
```

internal/drift/detect.go
```diff
@@ -0,0 +1,454 @@
...
+		detectEntry(key, funcName, typ, valuePair{crdValue, crdInterface}, valuePair{remoteValue, remoteInterface}, child)
+	}
+}
+
+func collectKeys(keyValues []reflect.Value, recipient map[string]reflect.Value) {
+	for _, keyValue := range keyValues {
+		if key, ok := asInterface(keyValue).(string); ok {
+			recipient[key] = keyValue
+		} else {
+			log.Panicf("map key must be of type string, got %T", asInterface(keyValue))
```
carlos-andres-osorio: `collectKeys` calls `log.Panicf` when a map key is not exactly of type `string`. `Detect` runs in the admission webhook request path, so this is a denial-of-admission vector.

## Evidence

This is more severe than "unexpected data shapes" implies — it's reachable via an **already-shipped, statically-typed field**, not just malformed input: `api/model/group/group.go:58` — `Member.Roles map[RoleScope]string` where `type RoleScope string`. `collectKeys` does `asInterface(keyValue).(string)`, which fails for any *named* string type (Go type assertions are exact-type, not kind-based).

```yaml
kind: Group
metadata:
  name: repro-rolescope-panic
spec:
  name: repro-rolescope-panic
  notifyMembers: true
  members:
    - source: gravitee
      sourceId: user@example.com
      roles:
        API: OWNER
```

```go
// real Group CRD, byte-identical crd/remote, no malformed input at all
result := drift.Detect(*crd.Spec.Type, *remote.Spec.Type)
```
```
crd.Spec.Members[0].Roles = map[group.RoleScope]string{"API":"OWNER"}
panic: map key must be of type string, got group.RoleScope
	internal/drift.collectKeys(...) detect.go:317
	internal/drift.detectIndexedMapItems(...)
	internal/drift.detectStruct(...)
	internal/drift.Detect(...)
```

The moment drift detection is wired up for `Group` resources, this panics on essentially any group with role assignments — the normal case, not an edge case.

internal/drift/detect.go
```diff
@@ -0,0 +1,454 @@
...
+	Interface any
+}
+
+// Detect detects the drift between two structs.
+// Goes recursively through the structs and detects the drift.
+// It uses the drift.EquivalenceFunc to determine if the two values are equivalent.
+// Each field is tagged with a drift.EquivalenceFunc name to determine which EquivalenceFunc to use.
+// By default, the EquivalenceFunc is reflect.DeepEqual.
+// The result is a tree of Result. It can be printed in a pseudo-yaml format.
+func Detect(crd any, remote any) Result {
```
carlos-andres-osorio, line 35: No test for a nil root on one side. `Detect` explicitly accepts nil roots (`assertRootIsStruct` skips nil, and `valuePairFromField` has a dedicated `v == nil` branch), but every test passes two non-nil structs; the panic table only covers non-struct roots.

## Evidence

Real CRD, `spec.endpointGroups[0].services.discovery`:

```yaml
kind: ApiV4Definition
metadata:
  name: repro-merge-fallback
spec:
  hrid: repro-merge-fallback
  type: PROXY
  listeners: []
  endpointGroups:
    - name: default-group
      services:
        discovery:
          enabled: false
        healthCheck:
          enabled: true
```

```go
crdServices := *crd.Spec.EndpointGroups[0].Services.Discovery // real v4.Service value
result := drift.Detect(crdServices, nil)
```
```
Detect(realCRDValue, nil) completed: DriftDetected=true
enabled: false != <nil>
overrideConfiguration: false != <nil>
```

Works correctly, no panic — confirming this is a real, intentionally-supported path with zero regression coverage in `detect_test.go`.

internal/drift/registry.go
```diff
@@ -0,0 +1,113 @@
...
+		log.Panicf("drift detection only compare non interface types")
+	}
+	if remoteType != nil && remoteType.Kind() == reflect.Interface {
+		log.Panicf("drift detection only compare non interface types")
+	}
+}
+
+// defaultStructEquivalence if one of the fields is exclusively nil, the whole struct is not equivalent and elements are skipped.
+// Otherwise, it is marked as cannot compare, but fields are not skipped.
+func defaultStructEquivalence(crd any, remote any) Equivalence {
```
carlos-andres-osorio, line 93: `defaultStructEquivalence`: no test for the XOR branch where an untagged pointer-to-struct field is nil on exactly one side, which marks the node `Inequivalent` with `Skip=true` and suppresses child detail. Every nested-nil test in `registry_test.go` uses fields tagged `drift:"empty-is-nil"`; the only untagged pointer-struct fixture field is never nil on one side in any test.

## Evidence

This is the same code path exercised by the `detect.go:429` finding's evidence in this review — real `v4.EndpointGroupServices` where `discovery` is present on the remote CRD but dropped from the re-applied CRD:

```yaml
# remote
endpointGroups:
  - name: default-group
    services:
      discovery:
        enabled: true
        overrideConfiguration: false
      healthCheck:
        enabled: true
        overrideConfiguration: false
```
```yaml
# crd
endpointGroups:
  - name: default-group
    services:
      healthCheck:
        enabled: true
        overrideConfiguration: false
```

That repro showed the `discovery` child gets silently dropped from the rendered tree. The gap here is specifically that no *permanent* test pins this branch.

internal/drift/equivalences.go
```diff
@@ -0,0 +1,216 @@
...
+	crdString, _ := crd.(string)
+	remoteString, _ := remote.(string)
+	if strings.TrimSpace(crdString) == strings.TrimSpace(remoteString) {
+		return Equivalence{Equivalent: Equivalent}
+	}
+	return Equivalence{Equivalent: Inequivalent}
+}
+
+// RFC3339 checks if the value is a valid RFC3339 time and if they represent the same time.
+func RFC3339(crd any, remote any) Equivalence {
```
carlos-andres-osorio, line 88: `RFC3339`: no test for the remote-side parse-error branch. Only the crd-side invalid timestamp is tested (`test/unit/drift/equivalence_test.go:335`, `drift.RFC3339("not-a-date", certStartsAt)`); every other equivalence function in the suite is deliberately tested in both directions, and this branch (`remoteRFC3339time, err := parseRFC3339(remoteString)`) is a distinct code path with its own error return that's never exercised.

test/unit/drift/equivalence_test.go
```diff
@@ -0,0 +1,401 @@
...
+	DescribeTable("should report equivalence",
+		func(crd, remote string) {
+			Expect(drift.RFC3339(crd, remote)).To(Equal(
+				drift.Equivalence{Equivalent: drift.Equivalent},
+			))
+		},
+		Entry("same RFC3339 timestamp", certStartsAt, certStartsAt),
+		Entry("same instant with different offset notation", certEndsAt, "2025-06-15T10:30:00Z"),
+		Entry("crd RFC3339 vs remote RFC3339Nano without fractional seconds", certStartsAt, "2024-06-15T10:30:00.000000000Z"),
+		Entry("crd RFC3339Nano vs remote RFC3339", "2024-06-15T10:30:00.123456789Z", "2024-06-15T10:30:00.123456789Z"),
```
carlos-andres-osorio: This entry is named "crd RFC3339Nano vs remote RFC3339" but passes the exact same string (`"2024-06-15T10:30:00.123456789Z"`) as both crd and remote:
```go
Entry("crd RFC3339Nano vs remote RFC3339", "2024-06-15T10:30:00.123456789Z", "2024-06-15T10:30:00.123456789Z"),
```
It never exercises a cross-format comparison — just confirms a string equals itself — so it's fake coverage of the mixed nano-vs-RFC3339 path it claims to test.

carlos-andres-osorio:
`ENABLE_DRIFT_DETECTION` defaults to `true` here (`internal/env/env.go:180`, `helm/gko/values.yaml`/`config.yaml`). This PR only adds the generic framework — no resource's `validateUpdate` calls `ValidateDrift` yet (that's #1714 and future resource PRs) — so by itself this default is inert. The concern is about the combination once resource wiring lands on top of it, since the default is set here.

**The behavior once wired:** GKO doesn't poll or watch APIM for out-of-band changes — reconciliation is purely Kubernetes-event-driven (CRD spec change, or a watched dependency like `ManagementContext`/templating Secrets changing). So if a resource is edited directly in APIM, GKO doesn't notice and doesn't correct it — not quickly, not eventually — unless something unrelated happens to trigger a reconcile for other reasons (in which case it's still silently overwritten, unaffected by this PR). The drift just persists, invisibly, until someone tries to `apply` an update to that resource, which is exactly when the new check fires.

**Why that's a compatibility break, not just a tightening:** before this, `apply` was a reliable, always-succeeding, idempotent operation — GitOps controllers and CI pipelines depend on that. After this default is active, the *same* `apply` can fail for reasons entirely unrelated to what's being changed (accumulated drift from anywhere, at any point in the resource's history), with:
- No way to discover ahead of an upgrade which resources are already drifted.
- No bundled remediation — nothing helps reconcile existing drift before the block kicks in.
- No opt-in step — every existing installation gets this on upgrade, since the default is enabled and there's no way to have set the opt-out annotation/env var before the feature existed.

That combination — a previously-reliable operation starting to fail unpredictably, for automated pipelines with no visibility into why — is what makes this feel like it needs a `feat!:` / `BREAKING CHANGE:` footer and a `CHANGELOG.md` entry on whichever PR first wires a resource to this default, rather than shipping quietly enabled.

Worth confirming:
- Was default-enabled a deliberate call, or should this ship default-*off* for its first release, with users opting in once they've had a chance to audit/reconcile existing drift?
- Is there (or should there be) a way to detect existing drift *before* upgrading, rather than users discovering it via a failed `apply`?

carlos-andres-osorio:
Follow-up to the earlier comment on default-enabled drift detection: there's a concrete interaction with GitOps self-heal (Argo CD, Flux) worth documenting.

GKO's reconcile loop writes `core.LastSpecHashAnnotation` onto the live object's `.metadata.annotations` on every spec-changing reconcile (`controllers/apim/application/application_controller.go:88`). That annotation is computed (a hash of the current spec) and isn't declared in a user's Git manifest — nor could it be, since its correct value changes with the spec and users have no way to compute it themselves.

Argo CD's default live-vs-desired diff is not scoped to `.spec` — any field present on the live object but absent from Git, including annotations, is compared and can trigger `OutOfSync`. This is a [documented, common pattern](https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/) for controller/webhook-added fields, and the standard fix is an explicit `ignoreDifferences` rule — annotations aren't excluded by default.

Put together: for any Argo CD Application with `selfHeal: true` (common, often-recommended) managing a GKO Application/Subscription, Argo CD will periodically detect the `LastSpecHashAnnotation` diff as drift and automatically attempt to correct it via a real `Update` to the resource — even though nothing about the actual desired spec changed. That automated, non-business-intent `Update` goes through the same `validateUpdate` → `drift.ValidateDrift` path as any other update. If APIM has any accumulated drift at that moment (unrelated to the annotation), the automated self-heal correction gets rejected too — turning what should be a harmless no-op reconciliation into a recurring, automated failure loop with no human in the loop to notice or fix it.

Recommend documenting (AGENTS.md and/or user-facing docs/Helm README):
- That `gravitee.io`-managed annotations like the spec-hash annotation are operator-owned and should be excluded from GitOps diffing via `ignoreDifferences` (or equivalent) to avoid spurious sync churn.
- That drift detection can reject such automated self-heal corrections when remote drift exists, and what the recommended remediation/opt-out path is for GitOps-managed resources.

main.go
benoitgravitee: Always call Init() as user may override

internal/admission/drift/validate.go
benoitgravitee: All admission controller already have compiled templates and report error and stop processing. I'll remove the extra compile, it is not necessary

internal/admission/drift/validate.go
benoitgravitee: Agreed. Will refactor later to avoid breaking all I already implemented in the other branch.

internal/admission/drift/validate.go
benoitgravitee: They will in the next PR.

benoitgravitee:
Was default-enabled a deliberate call, or should this ship default-off for its first release, with users opting in once they've had a chance to audit/reconcile existing drift?
It goes with the removal of read only mode. For now, users having drift may have used mAPI to update things and know that all will be erase on the next apply.
Now that they can change stuff in the console, that becomes "legal to do so" hence the drift detection from now on.

Is there (or should there be) a way to detect existing drift before upgrading, rather than users discovering it via a failed apply?
For the reasons above, there should not be a way to detect drift upon upgrade.

internal/drift/registry.go
```diff
@@ -0,0 +1,106 @@
...
+
+var equivalenceRegistry = &registry{
+	registry: make(map[registryEntry]EquivalenceFunc),
+}
+
+// Get returns the EquivalenceFunc function for the given kind.
+// If no function is registered for the given kind and name, it panics.
+// If no name is given (no drift tag), it returns a default function for the given kind:
+// - For slices and arrays, it returns a function that always returns CannotCompare
+// - For structs, it returns a function that returns Inequivalent if one of the fields is exclusively nil or returns CannotCompare
```
carlos-andres-osorio: The `Get` doc comment claims the default (no drift tag) struct behavior "returns Inequivalent if one of the fields is exclusively nil or returns CannotCompare." The actual default, `defaultStructEquivalence` (registry.go:93-98), unconditionally returns `Equivalence{Equivalent: CannotCompare, Skip: false}` — it never inspects nil-ness and never returns `Inequivalent`. That nil-vs-zero-value behavior belongs to the separately registered `EmptyIsNilStruct` function (`equivalences.go`, name `"empty-is-nil"`), not the untagged default path. Confirmed by `test/unit/drift/detect_test.go` ("reports drift when discovery is absent on crd but present on remote…", using `structWithTwoPtrStructNoDriftTags`): children are always compared for the default path regardless of nil-ness, never short-circuited to `Inequivalent`/skip. The same wrong claim is duplicated in the comment directly above the function at registry.go:91-92 and in the package doc at `internal/drift/doc.go:78-79`.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/core"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/drift"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/env"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// RefResolver is a function that resolves all references in the CRD;
```
carlos-andres-osorio: The `RefResolver` type doc says "it used on old and new CRD to compare the content that will be sent to the remote API," but `ValidateDriftWithContext` (validate.go:73) invokes `resolveRefs` exactly once, on `oldCopy` only. `newCopy` is passed straight into `mapDTO` at validate.go:91 without ever going through `resolveRefs`. This is a likely source of false-positive drift: when a user legitimately reconciles existing remote drift, O/R can be Inequivalent (resolved-old vs remote) while N/R is also Inequivalent purely because new refs are unresolved, so `evaluateDrift` reports drift and blocks an otherwise-valid apply. No test in `test/unit/drift/` or `test/unit/admission/drift/` exercises `ValidateDriftWithContext` to settle whether this asymmetry is intentional.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,115 @@
...
+	newCRD core.ContextAwareObject,
+	resolveRefs RefResolver,
+	getRemoteObject RemoteObjectGetter,
+	dtoMapper DTOMapper) *errors.AdmissionErrors {
+	return ValidateDriftWithContext(ctx, oldCRD, newCRD, func(ctx context.Context) (*apim.APIM, error) {
+		return apim.FromContextRef(ctx, newCRD.ContextRef(), newCRD.GetNamespace())
+	}, resolveRefs, getRemoteObject, dtoMapper)
+}
+
+func ValidateDriftWithContext(
```
carlos-andres-osorio, line 59: `ValidateDriftWithContext` (and its wrapper `ValidateDrift`) is new admission-gating logic with several distinct branches — drift-detection disabled short-circuit, `resolveRefs` producing a severe error, `resolveContext` failing, `getRemoteObject` producing a severe error, and the final drift-detected vs no-drift outcome — yet the diff adds zero tests for this file. This is the function that will actually reject or allow a Kubernetes admission request once wired into a resource's webhook, so a regression here would only surface once the framework is wired into a real CRD's admission path, and by then it fails invisibly (either silently disabling drift protection or wrongly blocking every valid update). All four collaborator functions (`RefResolver`, `RemoteObjectGetter`, `DTOMapper`, `ContextResolver`) are already expressed as injectable function types, so this is straightforward to unit test with fakes.

internal/drift/types.go
```diff
@@ -0,0 +1,198 @@
...
+	// evaluates if a drift occurred and return the result that will be used to display the drift if any.
+	result := evaluateDrift(or, nr)
+	// put it all in the merged result
+	merger.Equivalence = result.Equivalence
+	merger.Property = result.Property
+	merger.Index = result.Index
+	merger.CRDValue = result.CRDValue
+	merger.RemoteValue = result.RemoteValue
+
+	if len(or.Children) == len(nr.Children) {
```
carlos-andres-osorio: `Merge`'s central invariant, documented only in prose ("O/R and N/R have the same structure as they ran on the same struct"), is never checked in code. `mergeChildrenPositionally` (chosen whenever `len(or.Children) == len(nr.Children)`) zips children purely by index, without verifying that `orChild.Property`/`Index` match `nrChild.Property`/`Index`. Concretely: remote `{b:1}`, oldCRD `{a:1}`, newCRD `{c:1}` → `or` children `[a,b]` (alphabetically-sorted union), `nr` children `[b,c]` — both length 2, so positional pairing zips index 0 (`a` vs `b`) and index 1 (`b` vs `c`), silently mislabeling one field's drift as another's in the admission error message shown to users, with no error raised anywhere. `mergeChildrenByKey` already exists and pairs correctly by `childKey` but is skipped whenever the equal-length branch is taken first.

internal/drift/types.go
```diff
@@ -0,0 +1,198 @@
...
+	// Rest is no drift, so we return the result without drift to have a clean result (reason could be filled)
+	// Case 2
+	if nr.Equivalent == Equivalent {
+		return nr
+	}
+	// Case 4
+	if or.Equivalent == Equivalent {
+		return or
+	}
+	// should not happen but just in case
```
carlos-andres-osorio: `evaluateDrift` has an explicitly-acknowledged "should not happen but just in case" fallback branch, reached whenever the O/R and N/R equivalence statuses land on the one combination not otherwise handled (one side `CannotCompare`, the other `Inequivalent`). Instead of logging that an unexpected state occurred, it silently returns a zero-value `Result{}`, whose `Equivalent` field defaults to `CannotCompare` (0) — a status `Result.DriftDetected()` treats as "no drift." This is the opposite of the rest of the `drift` package's convention: every other unexpected/programmer-error condition (`detect.go`, `registry.go`) calls `log.Panicf` to fail loudly, but this one path fails silently and swallows what could be a real, reportable drift, with zero diagnostic trail. Given `Merge` is the exact function that decides whether an admission request is rejected for drift, silently mapping an unanticipated state to "no drift" means a real drift could pass admission undetected with no trace for anyone debugging why.

internal/admission/drift/validate.go
benoitgravitee: Tests will implemented in the next PR with actual drift detection on CRDs

internal/drift/types.go
benoitgravitee: Docs fixed

internal/drift/doc.go
carlos-andres-osorio: Looks addressed — the doc now reads "0-len or nil is considered as equivalent," which drops the misleading "unless lengths differ" claim from the earlier wording. Reopening just to confirm this was an intentional rewrite matching `EmptyIsNilLen`'s actual behavior, not an unrelated doc pass. Can you confirm?

internal/drift/detect.go
carlos-andres-osorio: Looks fixed — `handleStructField` now always appends a child result instead of writing onto the parent's `Equivalence`, and `defaultStructEquivalence` no longer has the XOR-nil-flagged `Inequivalent+Skip=true` branch at all (always returns `CannotCompare`/`Skip=false`), so the silent-loss scenario here can't happen anymore. Reopening to confirm this was the intended fix — note the same change also closes out the untested-branch complaint on registry.go:93, which I'm leaving resolved as the same fix rather than re-raising separately.

internal/drift/types.go
carlos-andres-osorio: Confirmed fixed — the fallback branch now calls `log.Panicf` instead of silently returning `Result{}`, matching the convention used elsewhere in the package. Reopening only to note this same fix also closes the earlier round's duplicate thread on this line (originally raised at line 156) — no separate action needed there.

helm/gko/templates/manager/config.yaml
carlos-andres-osorio: Confirmed fixed — `helm/gko/tests/manager/config_test.yaml` now has a "Should have drift detection disabled" case asserting `ENABLE_DRIFT_DETECTION == "false"` when `driftDetection.enabled: false` is set. Reopening just to close the loop / confirm this was intentional.

test/unit/drift/equivalence_test.go
carlos-andres-osorio: Looks addressed — the crd/remote values in that entry no longer match (previously identical strings), so the "RFC3339Nano vs RFC3339" label is now materially testing what it claims. Reopening just to confirm intentional.

internal/drift/registry.go
carlos-andres-osorio: Confirmed fixed — the doc comment on `Get` now says the struct default returns `CannotCompare`, matching `defaultStructEquivalence`'s actual behavior (previously claimed `Inequivalent`). Reopening just to close the loop / confirm intentional.

internal/drift/types.go
carlos-andres-osorio: Reopening — this is still present. Reproduced it directly this round with a real `v4.EndpointGroup` (map-typed `Headers` field): renaming a header key on the CRD side without touching remote gets merged positionally (`len(or.Children) == len(nr.Children)`) and reported as drift, because `mergeChildrenPositionally` zips children by index with no same-key check. The doc-comment fix noted here doesn't change `merge()`'s actual behavior — `mergeChildrenByKey` already exists and pairs correctly but is skipped whenever the equal-length branch is taken first.

internal/admission/drift/validate.go
carlos-andres-osorio: Reopening — confirmed still present this round. Real `*v1alpha1.ApiV4Definition` CRDs through the actual `ValidateDriftWithContext`: `resolveRefs` is still only invoked on `oldCopy` (validate.go:72); `newCopy` reaches `getRemoteObject`/`mapDTO` unresolved. Also traced two concrete downstream consequences via `evaluateDrift`'s case logic: (1) when both O/R and N/R disagree, the reported diff cites the raw unresolved placeholder rather than the real value — misleading admission errors; (2) when O/R already matches remote, the N/R side is silently discarded regardless of whether the new CRD's real (never-resolved) value would actually match remote — a masked exemption.

internal/drift/equivalences.go
carlos-andres-osorio: Reopening — still no test exercises the remote-side RFC3339 parse-error path end-to-end through `Detect()`/`Result.String()` (confirmed no fixture in test/unit/drift uses the rfc3339 equivalence tag). Same gap surfaces at the `formatValue` error-formatting branch (format.go:35) from the display side.

internal/drift/detect.go
```diff
@@ -0,0 +1,468 @@
...
+		// detect the item and pass it the right order, crd then remote
+		if crdIsLarger {
+			detectItem(property, i, leaderItem, followerItem, parent, ordered)
+		} else {
+			detectItem(property, i, followerItem, leaderItem, parent, ordered)
+		}
+	}
+}
+
+func detectItem(property string, i int, crdItem reflect.Value, remoteItem reflect.Value, parent *Result, ordered bool) {
```
carlos-andres-osorio, line 223: `detectItem` switches on `crdItem.Kind()` without dereferencing pointer elements first, unlike `detectStruct`/`detectEntry`, which both call `dereferenced()` before branching. For slice fields of pointer-to-struct type (`[]*T` — pervasive in `api/model`, e.g. `[]*Plan`, `[]*Endpoint`, `[]*VirtualHost`), `Kind()` is `reflect.Ptr`, so the `Struct`/`Map`/`Slice` cases are never reached and execution falls into the `default` branch, where the whole pointer is compared via `reflect.DeepEqual` — bypassing per-field traversal and every field-level equivalence tag (`trimmed`, `rfc3339`, `empty-is-nil`, ...) entirely.

## Evidence

```go
func TestPointerSliceEndpointNilVsEmptyTenantsFalsePositive(t *testing.T) {
	drift.Init()

	name := "backend-1"

	crd := v4.EndpointGroup{
		Name: "default-group",
		Endpoints: []*v4.Endpoint{
			{
				Name:    &name,
				Type:    string(v4.EndpointTypeHTTP),
				Tenants: nil,
			},
		},
	}

	remote := v4.EndpointGroup{
		Name: "default-group",
		Endpoints: []*v4.Endpoint{
			{
				Name:    &name,
				Type:    string(v4.EndpointTypeHTTP),
				Tenants: []string{},
			},
		},
	}

	result := drift.Detect(crd, remote)

	if result.DriftDetected() {
		t.Fatalf("expected NO drift (nil and empty Tenants are semantically equivalent), "+
			"but drift was detected.\nResult tree:\n%s", result.String())
	}
}
```

```
--- FAIL: TestPointerSliceEndpointNilVsEmptyTenantsFalsePositive (0.00s)
    empirical_test.go:78: expected NO drift (nil and empty Tenants are semantically equivalent), but drift was detected.
        Result tree:
        endpoints[0]: {0x1400029f350 http-proxy <nil> false <nil> <nil> <nil> <nil> []} != {0x1400029f350 http-proxy <nil> false <nil> <nil> <nil> <nil> []}
        headers: <nil> != <nil>
```

Real `v4.EndpointGroup`/`v4.Endpoint` CRD types pushed through the public `drift.Detect` entry point — the only difference is `Endpoints[0].Tenants`: `nil` (unset by a user's CRD) vs `[]string{}` (API-defaulted), semantically identical values that get reported as drift because the whole `*Endpoint` is diffed as an opaque struct dump instead of field-by-field. Verified interactively during review.

internal/drift/format.go
```diff
@@ -0,0 +1,79 @@
...
+		formatValue(this, b, indent)
+	}
+}
+
+func formatValue(this *Result, b *strings.Builder, indent int) {
+	addIndent(b, indent)
+	if this.Index != nil {
+		b.WriteString(fmt.Sprintf("%s[%d]: %v != %v", this.Property, *this.Index, resolve(this.CRDValue), resolve(this.RemoteValue)))
+	} else {
+		b.WriteString(fmt.Sprintf("%s: %v != %v", this.Property, resolve(this.CRDValue), resolve(this.RemoteValue)))
```
carlos-andres-osorio, line 40: A09 (Security Logging & Alerting Failures): the drift formatter writes every inequivalent field's CRD and remote values verbatim (`%s: %v != %v`) into `Result.String()`, which `internal/admission/drift/validate.go:95` embeds into a severe `AdmissionError` returned to the client — surfacing in `kubectl` output and the API server audit log. The compared DTOs are `api/model` types for APIM resources that routinely carry credentials (endpoint auth passwords, bearer tokens, header secrets). The only field-level escape hatches — `ignore`/`IgnoreSkip` — skip the comparison entirely, so there's currently no way to compare a sensitive field for drift while masking its value in the output. Any secret-bearing field that drifts leaks both the local and remote secret into the denial message and audit trail.

There's no existing convention in this codebase for marking a struct field as sensitive for drift comparison — `registry.go` dispatches purely on Go `reflect.Kind` plus an optional named `EquivalenceFunc`, with no redact/sensitive semantics. This looks like it needs a design decision (where the sensitivity marker would live, what gets printed for a redacted field, whether it needs to apply recursively to nested `Result.Children`, and whether the admission-response and audit-log paths need the same treatment) rather than a one-line patch.

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,113 @@
...
+func isDriftDetectionEnabled(newCRD runtime.Object) bool {
+	if coreObj, ok := newCRD.(core.Object); ok {
+		driftAnnot, hasAnnot := coreObj.GetAnnotations()[core.DriftDetectionAnnotation]
+		if hasAnnot && driftAnnot == env.TrueString {
+			return true
+		} else if hasAnnot && driftAnnot == env.FalseString {
+			return false
+		}
+	} else {
+		log.Panicf("CRD Does not implement core.Object interface: %T", newCRD)
```
carlos-andres-osorio: Nit: the "does not implement core.Object" defensive guard uses stdlib `log.Panicf` rather than the project's `internal/log` package used elsewhere in the codebase. The branch is effectively unreachable from `ValidateDrift` (whose params are `core.ContextAwareObject`, which extends `core.Object`), so this is a logging-convention inconsistency on a defensive guard, not a live bug.

internal/drift/types.go
```diff
@@ -0,0 +1,212 @@
...
+	// Case 2
+	if nr.Equivalent == Equivalent || nr.Equivalent == CannotCompare {
+		return nr
+	}
+	// Case 4
+	if or.Equivalent == Equivalent || or.Equivalent == CannotCompare {
+		return or
+	}
+
+	log.Panicf("Unable to evaluate drift between (Old vs. Remote) %v and %v (New vs. Remote)", or, nr)
```
carlos-andres-osorio, line 225: The new `log.Panicf("Unable to evaluate drift between...")` branch (replacing the old silent `return Result{}` fallback) has 0% test coverage. Enumerating all 9 `EquivalentStatus` pairs shows the first three `if` branches in `evaluateDrift` already exhaust every combination reachable through the only current production call site, so this branch is currently unreachable in practice — but `EquivalentStatus`, `Result.Equivalent`, and `Merge` are all exported, so an external test in `package drift_test` could construct a `Result` with an out-of-range `EquivalentStatus` and exercise the panic branch directly.

internal/drift/detect.go
carlos-andres-osorio: Confirmed with a reproduction test, not just by re-reading the diff.

## Evidence

Real `v2.Services`/`EndpointDiscoveryService` CRD types pushed through the public `drift.Detect` entry point, replicating this thread's exact scenario (remote has `discovery` configured, CRD drops it on re-apply):

```go
func TestDiscoveryDroppedIsNotSilentlyLost(t *testing.T) {
	remote := v2.Services{
		EndpointDiscoveryService: &v2.EndpointDiscoveryService{
			Service: &v2.Service{Enabled: true},
		},
		HealthCheckService: &v2.HealthCheckService{
			ScheduledService: &v2.ScheduledService{Service: &v2.Service{Enabled: true}},
		},
	}
	crd := v2.Services{
		HealthCheckService: &v2.HealthCheckService{
			ScheduledService: &v2.ScheduledService{Service: &v2.Service{Enabled: true}},
		},
	}

	result := drift.Detect(crd, remote)

	if !result.DriftDetected() {
		t.Fatalf("expected drift to be detected, got none")
	}
	if !strings.Contains(result.String(), "discovery") {
		t.Fatalf("REFUTED: dropped discovery field is absent from the result tree (silent loss):\n%s", result.String())
	}
}
```

```
--- PASS: TestDiscoveryDroppedIsNotSilentlyLost (0.00s)
    Result tree:
        discovery:
          enabled: <nil> != true
```

The fix works, though via a different mechanism than I first assumed: `defaultStructEquivalence` no longer has the XOR-nil `Inequivalent+Skip=true` branch at all — it now unconditionally returns `CannotCompare, Skip=false`, so untagged struct fields are never skipped by default and always recurse into per-field comparison. That also confirms the `registry.go:93` untested-branch complaint is the same code path, not a separate one — resolving this thread.

internal/drift/types.go
benoitgravitee: I find this a bit far fetched... I'd be in favour to leave it as is be defensive, just in case

internal/drift/format.go
benoitgravitee: I'll create a tech ticket for this one

internal/admission/drift/validate.go
```diff
@@ -0,0 +1,117 @@
...
+		return errs
+	}
+
+	apimClient, err := resolveContext(ctx)
+	if err != nil {
+		errs.AddSeveref("could not resolve context for CRD: %s", err.Error())
+		return errs
+	}
+
+	remoteObject, _ := getRemoteObject(apimClient, newCopy, errs)
```
carlos-andres-osorio: `remoteObject, _ := getRemoteObject(apimClient, newCopy, errs)` silently discards the `found` boolean that the `RemoteObjectGetter` contract explicitly returns to signal "the remote object was not found" (see the type doc a few lines up: "it returns the remote object and a boolean indicating if the object was found"). Nothing consumes that signal.

When a getter returns `(nil, false)` for a not-yet-created resource WITHOUT adding a severe error (not-found is a legitimate state, not an error), `errs.IsSevere()` is false and execution falls through to `drift.Detect(oldDTO, nil)` / `drift.Detect(newDTO, nil)`. Detecting against a nil remote compares every CRD field against a zero value, so a resource that simply does not exist remotely is relabelled as `drift detected:` and admission is rejected with a misleading diff instead of a clear "remote object not found" outcome.

The correctness of this path currently depends entirely on every `RemoteObjectGetter` implementation remembering to `AddSevere` on not-found — which the framework neither enforces nor logs.

test/unit/drift/detect_test.go
```diff
@@ -0,0 +1,1410 @@
...
+			crd := withAnyField{Value: "foo"}
+			remote := withAnyField{Value: 42}
+			Expect(func() {
+				drift.Detect(crd, remote)
+			}).To(PanicWith(MatchRegexp(`drift detection only work comparing values of same type`)))
+		})
+	})
+
+	Describe("named string map keys", func() {
+		It("compares without panicking", func() {
```
carlos-andres-osorio, line 1164: "named string map keys ... compares without panicking" calls `expectNoDrift(drift.Detect(member, member))` passing the same value as both crd and remote. Beyond the panic guard, this is a weak assertion — it never proves the named-string-key map path actually compares entries, so a regression that silently skipped maps keyed by a named string type (returning no children) would still pass this test.

The panic-guard intent is legitimate, but the compare-direction is untested — a companion case with differing values (e.g. crd `"OWNER"` vs remote `"USER"`) asserting the expected drift string would close the gap.

PR #1714, opened 2026-06-30
feat(GKO-2931): add drift detection for all resources

benoitgravitee:
## Issue

https://gravitee.atlassian.net/browse/GKO-2931

## Summary

History was rewritten so review can follow three layers instead of a single mixed stack:

1. **Drift framework** — comparison engine, admission glue (`internal/admission/drift`), Helm/env default-off, APIM equivalences, and framework tests.
2. **One commit per resource** — DTO, admission hook, unit/integration tests, and related service/example changes for that resource only (API v4, Application, Dictionary, Documentation, Group, Portal, PortalLink, PortalListing, SharedPolicyGroup, Subscription).
3. **Leftover consistency work** — concrete types in admission validators (`runtime.Object` → typed hooks, `ToReference` → `new()`), then Kind/CI/portal fixture alignment that landed alongside the drift work.

No functional change versus the previous tip of this branch; only commit boundaries changed.

## Test plan

- [ ] Review the framework commit in isolation (`feat(drift): update comparison engine and admission glue`)
- [ ] Review each resource commit as a self-contained DTO + admission + tests slice
- [ ] Skim the two leftover commits for signature/helper alignment vs Kind/CI/fixture noise
- [x] `make unit`
- [x] Integration admission suites for resources that gained drift detection

## Documentation In GKO Section in its page at the root level

# Drift detection

## What it does

When you manage Gravitee resources with GKO, the Kubernetes manifest is the source of truth. But nothing physically stops someone from changing the same API, application or dictionary directly in APIM — through the Console, the Management API, or the Automation API. When that happens the two sides silently diverge, and the next `kubectl apply` on the resource overwrites the APIM-side change without warning.

Drift detection closes that gap. It is an **admission-time guard**: when you update a GKO resource, the operator compares the live state in APIM with what the resource is supposed to look like, and refuses the update if APIM was changed behind the operator's back.

## How it works

Drift detection runs in the GKO admission webhook, on **update only**. Creations and deletions are never blocked by it.

On every update the operator:

1. Fetches the current state of the resource from APIM.
2. Compares it with **both** the previously applied version of your resource and the version you are applying now.
3. Reports drift only when the difference cannot be explained by your own change.

That third point is what makes the feature usable day to day. The comparison is three-way, so:

| Situation | Result |
|---|---|
| Nothing changed anywhere | allowed |
| You changed the manifest, APIM untouched | allowed |
| You changed the manifest to the same value someone set in APIM | allowed |
| APIM changed, your manifest did not | **drift** |
| You and APIM both changed, to different values | **drift** |

So if a colleague edits a description in the Console and you then update your manifest to that exact description, your update goes through: the manifest has caught up with reality. But if you try to push a different value, or reapply an unchanged manifest, the update is rejected and you are told exactly which fields diverged.

A few properties worth knowing:

- **The comparison is semantic, not textual.** It compares what the operator would actually send to APIM, not the raw YAML. Empty lists, unset optional fields, APIM-side defaults, whitespace-only differences, timezone-equivalent timestamps and case differences in enum-like values are treated as equal, so you don't get noise from fields you never set.
- **Server-managed fields are excluded.** Internal identifiers, cross-IDs and lifecycle/status metadata generated by APIM never count as drift.
- **Referenced values are resolved first.** Secrets, ConfigMaps and templated values are resolved on both sides before comparison, so an indirection is not reported as a difference.
- **It needs a Management Context.** Resources with no `contextRef` are not checked, since there is no APIM instance to compare against.
- **It needs the admission webhook.** If drift detection is enabled while the webhook is disabled, the operator refuses to start.

## Supported resources

| Resource | Drift detection |
|---|---|
| `ApiV4Definition` | supported |
| `Application` | supported |
| `Subscription` | supported |
| `SharedPolicyGroup` | supported |
| `Dictionary` | supported |
| `Portal` | supported |
| `PortalLink` | supported |
| `PortalListing` | supported |
| `Documentation` | supported |
| `Group` | supported only created with GKO 4.12 and above |
| `ApiDefinition` (v2) | not supported |

Unsupported resources ignore the setting entirely, including the annotation.

### Opt-in model

Drift detection is **disabled by default**, globally. You turn it on either globally, or per resource with an annotation.

The annotation always wins:

| `gravitee.io/drift-detection` on the resource | Global setting | Effective |
|---|---|---|
| `"true"` | anything | enabled |
| `"false"` | anything | disabled |
| absent (or any other value) | `false` (default) | disabled |
| absent (or any other value) | `true` | enabled |

```yaml
kind: Dictionary
metadata:
  name: shared-properties
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  # ...
```

The annotation is read from the resource you are applying, so you can flip it in the same `kubectl apply` that you want to let through — which is the standard escape hatch when you need to force an update over a known drift.

### Global settings

Helm values, with the equivalent environment variables:

| Helm value | Environment variable | Default | Meaning |
|---|---|---|---|
| `manager.driftDetection.enabled` | `DRIFT_DETECTION_ENABLED` | `false` | Master switch. When `false`, no comparison and no logs unless a resource opts in via annotation. Requires `manager.webhook.enabled`. |
| `manager.driftDetection.policy` | `DRIFT_DETECTION_POLICY` | `deny` | What to do when drift is found. |
| `manager.driftDetection.onRemoteMissing.policy` | `DRIFT_DETECTION_ON_REMOTE_MISSING` | `deny` | What to do when the resource does not exist in APIM (HTTP 404). |
| `manager.driftDetection.fetchFailurePolicy` | `DRIFT_DETECTION_FETCH_FAILURE_POLICY` | `deny` | What to do when the APIM state cannot be fetched (network error, 5xx, auth failure). |

All three policies accept the same values:

| Policy | Behaviour |
|---|---|
| `deny` | The update is rejected. `kubectl` shows an error with the list of differences. |
| `warn` | The update is accepted. `kubectl` shows a warning with the list of differences. |
| `allow` | The update is accepted. A short warning is written to the operator logs only, with no field-level detail. |

Policies are global; there is no per-resource policy annotation. The annotation only turns detection on or off.

Two notes on the secondary policies:

- `onRemoteMissing` set to `warn` or `allow` lets the operator recreate a resource that was deleted in APIM. Set to `deny`, you must fix APIM (or annotate the resource) before the update can proceed.
- Failures that happen *before* the comparison — a Management Context that cannot be resolved, or a Secret/ConfigMap reference that cannot be read — always reject the update, regardless of policy, because the operator cannot know whether there is drift or not.

## What the output looks like

When drift is detected the operator returns a message beginning with `drift detected:` followed by a compact tree of the differing fields. The **left value is the one from your Kubernetes resource, the right value is the one currently returned by the API**, separated by `!=`. Only differing fields are printed; everything that matches is omitted.

With the default `deny` policy, `kubectl` surfaces it as a rejected request:

```
Error from server: error when applying patch:
...
admission webhook "validate-apiv4definition.gravitee.io" denied the request:
drift detected:
description: "V4 Proxy API with extended fields" != "Edited from the console"
```

With `warn`, the same body is printed as `Warning:` and the update is applied.

The examples below use the fields of real CRDs.

### Simple scalar

An `Application` whose description was edited in APIM:

```
drift detected:
description: "Team A internal app" != "Team A internal app (do not use)"
```

A `Subscription` whose end date was pushed back in APIM:

```
drift detected:
endingAt: "2060-12-25T09:12:28Z" != "2070-11-28T12:20:00Z"
```

### Nested fields

Nesting is rendered as an indented tree; parents are printed only to give the path to the leaf that differs. An `ApiV4Definition` whose failover settings were changed:

```
drift detected:
failover:
  maxRetries: 5 != 2
  openStateDuration: 10000 != 30000
```

Deeper nesting, here an endpoint target changed in the Console:

```
drift detected:
endpointGroups[0]:
  endpoints[0]:
    configuration:
      target: "https://api.gravitee.io/echo" != "https://api.gravitee.io/echo/v2"
```

And a policy configuration inside a flow step:

```
drift detected:
flows[0]:
  request[0]:
    configuration:
      rate:
        limit: 10000 != 5000
```

### Arrays

List items are addressed by index. Lists of scalars print inline:

```
drift detected:
tags[0]: "internal" != "public"
tags[1]: "beta" != ""
```

An empty right-hand side means the item is absent in APIM; an empty left-hand side means it is absent from your resource.

Lists of objects print the index as a parent node:

```
drift detected:
properties[1]:
  value: "devex team" != "platform team"
```

A `PortalListing` where an API was moved in APIM:

```
drift detected:
apis[0]:
  location: "HEADER" != "FOOTER"
```

### Maps

Map entries are keyed by name and sorted, so output is stable between runs:

```
drift detected:
responseTemplates:
  CORS_PREFLIGHT_FAILED:
    text/html:
      status: 412 != 500
```

### Multiline strings

For fields such as page content or long descriptions, the two versions are diffed line by line and shown side by side. Unchanged runs are collapsed, and the `!=` marker appears once, on the first differing line:

```
drift detected:
content:  ... 1 identical line
          Version 1.                 !=  Version 2.
          Contact: team-a@example.com      Contact: platform@example.com
          ... 4 identical lines
```

### Long values

Values longer than 60 characters are clipped, with the number of hidden characters reported, so a single long description cannot flood your terminal:

```
drift detected:
plans[0]:
  description: "This plan requires an API Key to be set in the request heade..." (1 char hidden) != "Requires a subscription"
```

### Messages that are not field diffs

When the resource is missing in APIM, or cannot be fetched, there is no tree — just a single line:

```
remote [Application] [team-a/internal-app] not found during drift detection
```

```
failed to fetch remote [ApiV4Definition] [team-a/echo-api] during drift detection: 503 Service Unavailable
```

And with the `allow` policy, the only trace is in the operator logs:

```
drift detected for resource [Application] [team-a/internal-app], drift policy is 'allow': drift is ignored
```

## When is it worth enabling?

The value of drift detection depends on how easy it is to change a given resource outside the operator.

**The APIM Console already protects most core resources.** For `Subscription`, `Application`, `SharedPolicyGroup` and APIs, resources created by GKO are surfaced in the Console as Kubernetes-managed and are not editable there. The one caveat is APIs that have been **detached** in the Console: once detached, an API becomes editable and drift becomes possible again. So for these resource types, drift can realistically only be introduced API calls (another automation tool, a script, a Terraform provider) — a real risk, but a narrower one.

**Dictionaries and Portal resources are not protected the same way.** `Dictionary`, `Portal`, `PortalLink`, `PortalListing` and `Documentation` remain editable from the Console, and portal content in particular tends to attract manual edits from people who are not aware that it is managed declaratively. These are the resources where drift detection pays for itself, and where using the per-resource annotation is the recommended approach:

```yaml
kind: Documentation
metadata:
  name: getting-started
  annotations:
    gravitee.io/drift-detection: "true"
spec:
  # ...
```

A pragmatic rollout:

1. Leave the global switch off.
2. Annotate your dictionaries and portal resources with `gravitee.io/drift-detection: "true"`.
3. If you also want to catch out-of-band API-level automation, turn on the global switch — with `policy: warn` first, so you learn where drift already exists without blocking any pipeline, then move to `deny` once the noise is gone.

## Resolving a reported drift

You have three options, in order of preference:

1. **Adopt the APIM change.** Edit your manifest so the drifted fields match what the output shows on the right, then apply. Because the comparison is three-way, this is accepted.
2. **Discard the APIM change.** Revert the resource in APIM to the state your manifest describes, then reapply. The update then goes through untouched.
3. **Force the update.** Add `gravitee.io/drift-detection: "false"` to the resource in the same apply. Your manifest wins and the APIM-side change is overwritten. Remember to remove the annotation afterwards, otherwise the resource stays unprotected.


internal/apim/model/subscription.go
```diff
@@ -25,17 +25,50 @@ type SubscriptionRequest struct {
...
+	Key      string  `json:"key"`
+	ExpireAt *string `json:"expireAt,omitempty" drift:"rfc3339"`
+}
+
+type SubscriptionDTO struct {
+	ID                    string                             `json:"id"`
+	ApiID                 string                             `json:"apiId"`
+	AppID                 string                             `json:"applicationId"`
+	PlanID                string                             `json:"planId"`
+	Status                string                             `json:"status"`
```
carlos-andres-osorio: `SubscriptionDTO.Status` has no `drift:"ignore"` tag, unlike the analogous `ApplicationDTO.Status` (which is tagged `ignore`). The subscription `dtoMapper` never sets `Status` on either the old or new local DTO, but both remote getters populate it (e.g. `ACCEPTED`). So every comparison sees CRD `Status==""` vs remote `Status=="ACCEPTED"`, which is `Inequivalent` under default `reflect.DeepEqual` on BOTH old-vs-remote and new-vs-remote, which `drift.Merge` reports as drift.

Since drift detection is enabled by default, this would reject every Subscription CRD update the moment this PR merges — not just drifted ones, all of them, including completely benign spec-only changes. The integration test's `assert.DriftDetected` only substring-matches the expected message, so it can't catch this extra spurious drift line.

internal/apim/model/subscription.go
```diff
@@ -25,17 +25,50 @@ type SubscriptionRequest struct {
...
+	StartingAt            string                             `json:"startingAt" drift:"rfc3339"`
+	EndingAt              string                             `json:"endingAt" drift:"rfc3339"`
+	Metadata              map[string]string                  `json:"metadata,omitempty" drift:"empty-is-nil"`
+	ApiKeys               []ApiKeySpec                       `json:"apiKeys,omitempty" drift:"empty-is-nil"`
+	ConsumerConfiguration subscription.ConsumerConfiguration `json:"consumerConfiguration,omitempty"`
```
carlos-andres-osorio: `SubscriptionDTO.ConsumerConfiguration` (and `AutomationSubscriptionDTO.ConsumerConfiguration`) changed from `*subscription.ConsumerConfiguration` to a value type while keeping `json:"consumerConfiguration,omitempty"`. `omitempty` is a no-op on non-pointer struct values, so every subscription import now serializes `"consumerConfiguration":{"entrypointId":""}` — a field that was previously omitted entirely for the common non-push-plan case — to the automation API, which can fail remote validation or overwrite the remote consumer configuration.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
...
+	return func(ctx context.Context) (*apim.APIM, error) {
+		return apim.FromContextRef(ctx, app.ContextRef(), app.GetNamespace())
+	}
+}
+
+func dtoMapper(api core.ApiDefinitionObject, app core.ApplicationObject, plan core.PlanModel) drift.DTOMapper {
+	return func(crd any) any {
+		sub, _ := crd.(*v1alpha1.Subscription)
+		if k8s.IsAutomationAPIManaged(sub) {
+			apiNSName := refs.NewNamespacedName(api.GetName(), api.GetNamespace())
```
carlos-andres-osorio: `dtoMapper` builds HRIDs with swapped arguments (`refs.NewNamespacedName(api.GetName(), api.GetNamespace())` against a `NewNamespacedName(namespace, name string)` signature), so the local DTO carries `name-namespace` HRIDs while the remote subscription carries `namespace-name`. Since both old and new local DTOs are wrong the same way, `drift.Merge` marks `id`/`apiId`/`applicationId` as drifted on every update of an automation-managed subscription whenever name != namespace, falsely blocking updates.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
...
+	return nil
+}
+
+func resolveContext(app core.ContextAwareObject) drift.ContextResolver {
+	return func(ctx context.Context) (*apim.APIM, error) {
+		return apim.FromContextRef(ctx, app.ContextRef(), app.GetNamespace())
+	}
+}
+
+func dtoMapper(api core.ApiDefinitionObject, app core.ApplicationObject, plan core.PlanModel) drift.DTOMapper {
```
carlos-andres-osorio: `dtoMapper` never sets `Status` on the local `SubscriptionDTO`. Combined with the missing `drift:"ignore"` tag on `SubscriptionDTO.Status`, old and new local DTOs both compare `"" != "ACCEPTED"` against the remote, so `drift.Merge` reports drift on `status` for every subscription update while drift detection is enabled.

controllers/apim/subscription/internal/update.go
```diff
@@ -67,9 +66,9 @@ func CreateOrUpdate(ctx context.Context, subscription *v1alpha1.Subscription) er
 		sub.EndingAt = *spec.EndingAt
 	}
 	for _, k := range spec.ApiKeys {
-		sub.ApiKeys = append(sub.ApiKeys, submodel.AutomationApiKeySpec(k))
+		sub.ApiKeys = append(sub.ApiKeys, model.ApiKeySpec(k))
 	}
-	sub.ConsumerConfiguration = spec.ConsumerConfiguration
+	sub.ConsumerConfiguration = utils.SafeDereference(spec.ConsumerConfiguration)
```
carlos-andres-osorio: `sub.ConsumerConfiguration = utils.SafeDereference(spec.ConsumerConfiguration)` regresses the Import payload: the field changed from a pointer (nil ⇒ omitted from JSON) to a value struct with an ineffective `omitempty`, so every subscription import — including all non-push subscriptions with no consumer configuration — now PUTs `"consumerConfiguration":{"entrypointId":""}` to the automation API instead of omitting the field.

internal/apim/service/subscriptions.go
```diff
@@ -81,26 +80,26 @@ func (svc *Subscriptions) Subscribe(apiID, appID, planID string) (*model.Subscri
...
 
 // GetByHRIDWithLegacyAPI For tests purposes only.
 func (svc *Subscriptions) GetByHRIDWithLegacyAPI(
-	apiUUID, subscriptionHRID string) (*subscription.AutomationSubscription, error) {
+	apiUUID, subscriptionHRID string) (*model.SubscriptionDTO, error) {
 	url := svc.AutomationTarget("apis").WithPath(apiUUID).
 		WithPath("subscriptions").WithPath(subscriptionHRID).WithQueryParam("hridContainsApiUUID", "true")
 
-	sub := new(subscription.AutomationSubscription)
+	sub := new(model.SubscriptionDTO)
```
carlos-andres-osorio: `GetByHRIDWithLegacyAPI` decodes the automation-API response directly into `model.SubscriptionDTO` (json tags `id`/`apiId`/`applicationId`/`planId`), but the endpoint returns the hrid-shaped payload (`hrid`/`apiHrid`/`applicationHrid`/`planHrid`) — the same endpoint `GetByHRID` (five lines above) correctly decodes into `AutomationSubscriptionDTO` and converts with `.ToLegacy()`. As written, `ID`, `ApiID`, `AppID` and `PlanID` are silently left empty.

internal/admission/application/validate.go
```diff
@@ -215,6 +227,34 @@ func validateDryRun(ctx context.Context, app core.ApplicationObject) *errors.Adm
...
 
+func resolveAppRefs(ctx context.Context, o runtime.Object) error {
+	app, ok := o.(*v1alpha1.Application)
+	if !ok {
+		return nil
+	}
+	return appResolve.ResolveClientCertificates(ctx, app.Spec.Settings, app.GetNamespace(), app.GetName())
+}
+
+func getRemoteApp(apimClient *apim.APIM, o runtime.Object, errs *errors.AdmissionErrors) (any, bool) {
```
carlos-andres-osorio: `getRemoteApp` converts any fetch error, including not-found, into a severe admission error. For an Application that was never successfully synced (`Status.ID == ""`), `PopulateIDs` unconditionally sets `Spec.HRID` from name/namespace, so `getRemoteApp` always takes the `GetByHRID` branch — and a legitimate 404 (no remote counterpart exists yet) is indistinguishable from a transport failure. Every subsequent spec update is rejected while drift detection is enabled — including a corrective update to fix a resource whose initial definition was wrong and never synced in the first place.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
...
+				admissionErrors.AddSeveref("cannot fetch Subscription during drift detection from Api HRID %s and Subscription HRID %s: %s",
+					apiNSName.HRID(),
+					subNSName.HRID(),
+					err.Error())
+				return nil, false
+			}
+			return *remoteSub, true
+		}
+		crdSub, _ := object.(*v1alpha1.Subscription)
+		remoteSub, err := apimClient.Subscription.GetByID(api.GetID(), crdSub.Status.ID)
```
carlos-andres-osorio: Same never-synced gap on the legacy path: for a subscription without the `AutomationAPIManaged` condition and `Status.ID == ""`, `GetByID(api.GetID(), "")` is called with a literally empty subscription ID, which cannot succeed structurally; the resulting error is added as severe, blocking every update to a not-yet-synced subscription — including the very fix a user needs to submit.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
...
+
+		return *remoteSub, true
+	}
+}
+
+func resolveRefs(context.Context, runtime.Object) error {
+	return nil
+}
+
+func resolveContext(app core.ContextAwareObject) drift.ContextResolver {
```
carlos-andres-osorio: `resolveContext(app)` resolves the APIM client from the Application's `ContextRef`, but the subscription controller performs the actual sync using the API's context (`apim.FromContextRef(ctx, api.ContextRef(), ns)`). If the Application and the API reference different ManagementContexts, the drift check queries an APIM environment where the subscription does not exist, producing a spurious severe "cannot fetch Subscription" error.

Note `validateContextRefs` (which rejects exactly this mismatch — "management contexts must match between application and API") is only invoked in `validateCreate`, never in `validateUpdate` — so nothing else in this file catches a context divergence introduced after creation (e.g. the Application's `contextRef` edited later).

internal/admission/application/validate.go
```diff
@@ -215,6 +227,34 @@ func validateDryRun(ctx context.Context, app core.ApplicationObject) *errors.Adm
 	return errs
 }
 
+func resolveAppRefs(ctx context.Context, o runtime.Object) error {
```
carlos-andres-osorio: `resolveAppRefs` returns `nil` (success) when the passed object is not a `*v1alpha1.Application` (`if !ok { return nil }`). This silently signals "references resolved" to `ValidateDriftWithContext`, which then compares unresolved certificate references/templates against the remote — an unexpected type silently skips certificate reference resolution instead of aborting.

carlos-andres-osorio: `resolveAppRefs` inlines Secret/ConfigMap-backed client certificate content into the DTO before comparison, so on drift the full resolved `Content` — which the CRD deliberately kept out of the spec via a Secret ref — is printed verbatim in the admission error via `drift.Result.String`, leaking it to users who have update rights on Applications but no read rights on the referenced Secret.

internal/admission/application/validate.go
carlos-andres-osorio: The legacy branch of `getRemoteApp` (`Spec.HRID == ""` → fetch by `Spec.ID` via `Applications.GetByID`, the pre-4.11/4.12 path) is never exercised by any test; the only drift test always takes the HRID/`GetByHRID` branch.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
 	}
 	return nil
 }
+
+func remoteSubscriptionGetter(api core.ApiDefinitionObject) drift.RemoteObjectGetter {
```
carlos-andres-osorio: The non-automation-managed (legacy `GetByID`/UUID DTO) path of `remoteSubscriptionGetter`/`dtoMapper` is untested; the only integration test imports the subscription in a way that always takes the automation/HRID branch.

internal/admission/application/validate.go
```diff
@@ -215,6 +227,34 @@ func validateDryRun(ctx context.Context, app core.ApplicationObject) *errors.Adm
...
+	if app.Spec.HRID != "" {
+		remoteApp, err := apimClient.Applications.GetByHRID(app.Spec.HRID)
+		if err != nil {
+			errs.AddSeveref("cannot fetch Application during drift detection from HRID %s: %s", app.Spec.HRID, err.Error())
+			return nil, false
+		}
+		return *remoteApp, true
+	}
+	// prior to 4.11 resources
+	remoteApp, err := apimClient.Applications.GetByID(app.Spec.ID)
```
carlos-andres-osorio: Read/write API asymmetry in the legacy application drift path: writes always go through the Automation API even for pre-4.11/4.12 apps, but drift fetches legacy apps via the v1 Management API (`GetByID`), which does not return `metadata`/`members`/`notifyMembers` inline while `ApplicationSpec.ToDTO` populates all three from the spec — both comparisons diverge from remote identically, so any legacy application declaring metadata, members, or notifyMembers can never pass update admission.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
...
+				ID:                    subNSName.HRID(),
+				ApiID:                 apiNSName.HRID(),
+				AppID:                 appNSName.HRID(),
+				PlanID:                sub.Spec.Plan,
+				StartingAt:            sub.Status.StartedAt,
+				EndingAt:              utils.SafeDereference(sub.Spec.EndingAt),
+				Metadata:              sub.Spec.Metadata,
+				ApiKeys:               mapApiKey(sub.Spec.ApiKeys),
+				ConsumerConfiguration: utils.SafeDereference(sub.Spec.ConsumerConfiguration),
+			}
```
carlos-andres-osorio: Drift comparison includes secret material — `ApiKeySpec.Key` and `ConsumerConfiguration.EntrypointConfiguration` (which can carry basic-auth credentials) — and `drift.Result.String` renders raw `crd != remote` values into the admission error, which is echoed by kubectl and typically captured in API-server audit logs and CI output.

test/unit/drift/application_detect_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/application"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim/model"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/drift"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+var _ = Describe("Application Drift detection", func() {
+
+	DescribeTable("equivalent values",
```
carlos-andres-osorio, line 26: The new DTO drift unit table contains only "equivalent values" entries — there is no entry asserting that a genuinely different, comparable DTO field IS reported as drift. If the drift tags on `ApplicationDTO` regressed to skip/always-equivalent for every field, all of these unit tests would still pass; the framework-level `detect_test.go` covers generic inequivalence but not the per-DTO tag wiring.

test/unit/drift/subscription_detect_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/utils"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim/model"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/drift"
+	. "github.com/onsi/ginkgo/v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var _ = Describe("Application Drift detection", func() {
+
+	DescribeTable("equivalent values",
```
carlos-andres-osorio, line 25: Same one-sided gap for `SubscriptionDTO`: the table asserts only `expectNoDrift`; no entry asserts drift for a genuinely different tagged field (`rfc3339` dates, `empty-is-nil` collections, `unstructured` consumer config), so a mis-tag would not be caught here.

test/integration/admission/application/update_withContext_drift_test.go
```diff
@@ -0,0 +1,66 @@
+package application
```
carlos-andres-osorio, line 1: Both new integration spec files (this one and `test/integration/admission/subscription/update_withContext_drift_detect.go`) are missing the `_test.go` suffix, violating the documented convention. As plain `.go` files they compile into the non-test package (pulling ginkgo/gomega into the regular build graph) and are exempt from test-file lint rules; the two names are also mutually inconsistent (`drift_detect` vs `drift_detected`).

test/integration/admission/subscription/update_withContext_drift_detect_test.go
```diff
@@ -0,0 +1,76 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
```
carlos-andres-osorio: Same missing `_test.go` suffix as the application-side file; every sibling spec in the directory uses `_test.go`.

test/integration/admission/subscription/update_withContext_drift_detect_test.go
```diff
@@ -0,0 +1,76 @@
...
+		WithContext(constants.ContextWithCredentialsFile).
+		Build()
+
+	clientID := random.GetName()
+	fixtures.Application.Spec.Settings.App.ClientID = &clientID
+	fixtures.Subscription.Namespace = constants.Namespace
+
+	fixtures.Apply()
+
+	It("should fail if the ending date is changed on the remote", func() {
```
carlos-andres-osorio: `remoteSubscriptionGetter`/`dtoMapper` branch on `k8s.IsAutomationAPIManaged` (HRID/automation vs legacy Status.ID via Management API), but the single drift scenario in this file exercises only the automation branch — the legacy ID-based fetch/DTO path ships with no integration coverage.

internal/apim/model/subscription.go
```diff
@@ -25,17 +25,50 @@ type SubscriptionRequest struct {
...
+}
+
+type AutomationSubscriptionDTO struct {
+	HRID                  string                             `json:"hrid"`
+	ApplicationHrid       string                             `json:"applicationHrid"`
+	PlanHrid              string                             `json:"planHrid"`
+	ApiHrid               string                             `json:"apiHrid"`
+	Status                string                             `json:"status"`
```
carlos-andres-osorio: `SubscriptionDTO` and `AutomationSubscriptionDTO` are near-identical parallel structs bridged by hand-written field-by-field converters with no compile-time link; adding a field to one without updating both converters silently drops it from the round-trip and from drift comparison.

internal/admission/drift/validate.go
```diff
@@ -39,6 +39,13 @@ type RemoteObjectGetter func(*apim.APIM, runtime.Object, *errors.AdmissionErrors
 // DTOMapper is a function that converts the CRD into a DTO that can be compared with the remote object.
 type DTOMapper func(any) any
 
+// MapDTO wraps a typed mapper as a DTOMapper.
+func MapDTO[T any, D any](mapper func(T) D) DTOMapper {
```
carlos-andres-osorio, line 39: `MapDTO` performs an unchecked assertion `mapper(o.(T))`; if the framework ever passes an object whose concrete type is not `T`, this panics inside an admission webhook rather than surfacing a controlled `AdmissionError`.

internal/admission/application/validate.go
```diff
@@ -215,6 +227,34 @@ func validateDryRun(ctx context.Context, app core.ApplicationObject) *errors.Adm
...
+func resolveAppRefs(ctx context.Context, o runtime.Object) error {
+	app, ok := o.(*v1alpha1.Application)
+	if !ok {
+		return nil
+	}
+	return appResolve.ResolveClientCertificates(ctx, app.Spec.Settings, app.GetNamespace(), app.GetName())
+}
+
+func getRemoteApp(apimClient *apim.APIM, o runtime.Object, errs *errors.AdmissionErrors) (any, bool) {
+	app, _ := o.(*v1alpha1.Application)
```
carlos-andres-osorio: `getRemoteApp` does `app, _ := o.(*v1alpha1.Application)` and immediately calls `app.PopulateIDs(...)`; a failed assertion yields a nil pointer and panics the admission webhook. Depending on `failurePolicy`, repeated panics can turn the validating webhook into a fail-open path or a denial of service on Application updates.

internal/admission/subscription/validate.go
```diff
@@ -424,3 +441,89 @@ func validateEndingAt(endingAt *string) *errors.AdmissionError {
 	}
 	return nil
 }
+
+func remoteSubscriptionGetter(api core.ApiDefinitionObject) drift.RemoteObjectGetter {
+	return func(apimClient *apim.APIM, object runtime.Object, admissionErrors *errors.AdmissionErrors) (any, bool) {
+		condAwareSub, _ := object.(core.ConditionAwareObject)
```
carlos-andres-osorio: `remoteSubscriptionGetter` and `dtoMapper` both discard the `ok` from their type assertions; on any mismatch these propagate `nil` into `IsAutomationAPIManaged`/`sub.Spec.*`, causing a nil-deref panic rather than a surfaced admission error.

carlos-andres-osorio:
> **api/model/subscription/subscription.go:119**

`AutomationApiKeySpec` and `AutomationSubscription` (line 124) are now dead code: their only remaining references are each other and the generated deepcopy; all runtime usages were replaced by `model.ApiKeySpec` and `model.AutomationSubscriptionDTO`.

test/unit/drift/subscription_detect_test.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/subscription"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/utils"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/apim/model"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/internal/drift"
+	. "github.com/onsi/ginkgo/v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var _ = Describe("Application Drift detection", func() {
```
carlos-andres-osorio, line 23: `Describe("Application Drift detection", ...)` mislabels the subscription suite as "Application" (copy-paste from `application_detect_test.go`); a failure in this suite reports under the wrong name, misdirecting whoever triages it.

test/integration/admission/application/update_withContext_drift_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/assert"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/constants"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/fixture"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/test/internal/integration/labels"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
```
carlos-andres-osorio, line 17: The Apache license header is placed after the import block instead of at the top of the file, unlike every other source file in the repo; `make lint-fix`/addlicense conventions expect it as the leading comment.

internal/apim/model/subscription.go
benoitgravitee: status is not compared

helm/gko/values.yaml
```diff
@@ -201,8 +201,8 @@ manager:
     ## @param manager.templating.enabled If `false` string [templating](https://documentation.gravitee.io/gravitee-kubernetes-operator-gko/guides/templating) is not evaluated on resources and template markers will remain untouched.
     enabled: true
   driftDetection:
-    ## @param manager.driftDetection.enabled If `false`, drift detection is disabled globally. Individual resources can override it via the `gravitee.io/drift-detection` annotation.
-    enabled: true
+    ## @param manager.driftDetection.enabled If `true`, drift detection is enabled globally. Individual resources can override it via the `gravitee.io/drift-detection` annotation.
```
a-cordier:
```suggestion
    If true, admission webhooks deny updates to resources whose remote APIM state has drifted from the last applied spec. Requires manager.webhook.enabled. Applies to Automation-API-managed resources; Portal, Documentation, PortalListing and PortalLink additionally require a per-resource opt-in. Individual resources can opt in or out via the gravitee.io/drift-detection: "true"|"false" annotation.
```

hack/scripts/run-kind.mjs
```diff
@@ -43,6 +43,7 @@ const MONGO_IMAGE_TAG = await Mongo.getImageTag();
```
a-cordier, line 43: why do we update kind setup here ? just curious ...

helm/gko/values.yaml
a-cordier: I'm actually thinking of giving more control to the user on that

```yaml
manager:
  driftDetection:
    mode: "off"          # off | warn | enforce
    onRemoteMissing: "enforce"  # enforce | warn | allow
    failurePolicy: fail  # fail | ignore
```

I don't think it's a big deal to add WDYT ?

hack/scripts/run-kind.mjs
benoitgravitee: I want DCR support to test all kinds of app and plans and oauth apps require a DCR, hence a license

helm/gko/values.yaml
benoitgravitee:
```YAML
manager:
  driftDetection:
    
    # If true, admission webhooks can deny updates to resources whose remote state has drifted from the last applied spec.
    # Requires manager.webhook.enabled. controller will not start if disabled.
    # It occurs after other validations have been performed.
    # V2 APIs are not supported.
    # Is overriden by CRD level via annotation gravitee.io/drift-detection: "true"|"false"
    # If false everything is bypassed and no logs are produced.
    enabled: false
    
    # Policy when drift is detected:
    # deny: blocks the update with kubectl error message displaying the differences
    # warn: displays a kubectl warning message displaying the differences but allows the update
    # allow: just logs a simple warn message the controller logs only.
    policy: deny # warn | allow
    
    # When the admission controller does not find the resources on the remote state (404)
    onRemoteMissing:
    
      # Same as above except not differences are logged.
      # Be aware that warn and allow policies will let the controller recreate the resource on the remote.
      policy: deny # warn | allow
    
    # If failure occurs while fetching the remote state (non-404 errors or network):
    # deny: blocks the update with an error message.
    # warn: displays a kubectl warning but allow the update.
    # allow: just logs the error as a warning in the controller logs only.
    fetchFailurePolicy: "deny"    # warn | allow
```

benoitgravitee: Here is my proposal

a-cordier: 👍 but should fetchFailurePolicy: "fail"    # warn | ignore be at the driftDetection level rather than just declared for onRemoteMissing ?

in that case maybe a onFetchFailure.policy (like onRemoteMissing)

helm/gko/values.yaml
```diff
@@ -201,8 +201,15 @@ manager:
...
+    enabled: false
+    ## @param manager.driftDetection.policy Policy when drift is detected: `deny` blocks the update with a kubectl error message displaying the differences; `warn` displays a kubectl warning with the differences but allows the update; `allow` logs a simple warning in the controller logs only.
+    policy: deny
+    onRemoteMissing:
+      ## @param manager.driftDetection.onRemoteMissing.policy Policy when the admission controller does not find the resource on the remote (HTTP 404). Same values as `manager.driftDetection.policy` (`deny`, `warn`, or `allow`), except no differences are logged. `warn` and `allow` let the controller recreate the resource on the remote.
+      policy: deny
+    ## @param manager.driftDetection.fetchFailurePolicy Policy if fetching the remote state fails (non-404 errors or network): `deny` blocks the update with an error message; `warn` displays a kubectl warning but allows the update; `allow` logs the error as a warning in the controller logs only.
+    fetchFailurePolicy: deny
```
a-cordier: I would maybe go with `onFetchFailure.policy` for that one, this way it's open to extension (who knows) and it follows the `onRemoteMissing.policy `naming pattern WDYT ?

a-cordier (approved):

a-cordier (approved):

PR #1721, opened 2026-07-10
fix: comply to automation API for legacy API

benoitgravitee:
## Issue
https://gravitee.atlassian.net/browse/GKO-3042
## Description
Regression introduced in 4.12.4


tenshinhigashi (approved):

ctschacher:
Validated this fix end to end on the local upgrade survival lane (APIM 4.11 + released GKO 4.11.10, upgraded in place to APIM 4.12.x-latest + an operator built from this branch): carried-over APIs with spaced plan/page keys and lowercase flow mode reconcile after the resync, accept updates with unchanged identifiers, and delete cleanly. All 5 after-phase tests pass, including the new non-HRID scenario from GKO-3041 that first caught GKO-3042. The same lane against master fails deterministically at the admission dry-run.

ctschacher (approved):

PR #1760, opened 2026-08-03
test: consolidate the platform-test suite onto one authoring model

ctschacher:
## Purpose

The e2e suite had four ways to author a test and six ways to invoke it. The provisioner layer from the [spike-provisioners](https://gravitee.atlassian.net/wiki/x/AgBFFQ) doc landed in #1753, but the adoption step never happened: `ProvisionerView` had one call site, no journey used it, and shared bodies still branched with `isGko()`.

This is PR 1 of 2 and covers the framework. PR 2 migrates the ~100 migratable `tests/gko/` tests into journeys.

## Summary of changes

**One authoring model.** `assertProvisioner` now polls — it read once, while `read()` is specified to *throw* until it can tell, so it raced. Adopted after `remove()` and on failure paths. Deliberately **not** after `update()`: GKO's condition `observedGeneration` can lag indefinitely (GKO-2940), so a post-update read returns the pre-update `Accepted=True` and asserts nothing, while gating on `observedGeneration` would reintroduce the same hang `kubectl wait` suffers. With the view in place `GkoChecks` had nothing left, so it and `isGko` are gone; `checks` is now optional and Terraform-only.

**Lane selection is tag-only.** Dropping `testDirSegment`/`testIgnore` untied lane membership from the folder tree. The failure modes are silent — a missing tag runs a test in *every* lane, a duplicate tag in none, and both still pass an untagged full run — so `npm run check:lanes` asserts the lanes partition the suite exactly. It needs no cluster, and I verified it fails on an untagged test.

**One runner.** The upgrade specs no longer bypass `scripts/e2e.mjs`; they are `--suite upgrade --phase before|after`, with the npm scripts as aliases so CI is unchanged.

**Journeys grouped by persona** (`api-producer` / `api-consumer` / `platform-admin`). Applying the rule caught a mis-named journey: `consume-message-api` consumes nothing — its own README said "As an event API producer" — now `publish-a-message-api`. Group rename + delete were tested twice with assertions that had drifted apart; both are folded into `create-group-with-member`, which now asserts the invariant neither side covered (a rename updates in place, so the id survives).

**Removals.** The Chainsaw-era `platform-test` CLI (no caller, never published); its config loader moves to `src/config/`. `runner-examples` refreshed — they predated the provisioner layer — plus a README and a typecheck so they stop rotting.

**PARITY.md was wrong.** It claimed the provider exposes "exactly 6 standalone resource types"; on `origin/main` it exposes 10, and the inline set was understated (`apim_apiv4.members`, `.console_notification`, `.metadata`; `apim_application.settings.tls.client_certificate`). The provider is Speakeasy-generated from `automation-api-oas.yaml` and maps 1:1 onto the Automation API paths, which GKO also writes through — so "no Terraform resource" means "no Automation API path yet", a backlog item rather than a hard limit. That roughly doubles what PR 2 can migrate. Rewritten with a regeneration recipe, and its job split from AGENTS.md.

Verification: full suite green (357 passed / 6 skipped); lanes partition exactly; all 8 commits typecheck independently. Three full runs (two on this branch, one on `master`) each failed a *different* non-reproducible test — the gateway is `OOMKilled` once per run at its 512Mi default, filed as GKO-3082.

## Out of scope

- PR 2: migrating the ~100 migratable `tests/gko/` tests into ~22 journeys.
- Portal journeys — `apim_portal`, `apim_portal_listing`, `apim_documentation_portal` have no GKO CRD.
- Event-API consumption coverage — no test actually receives messages today.
- The gateway 512Mi OOM (GKO-3082).

See: https://gravitee.atlassian.net/browse/GKO-3081


a-cordier (approved):

ctschacher:
A code review by a different LLM (Codex) was done locally but didn't surface any issues.

PR #1772, opened 2026-08-11
feat: support portal homepage area on Documentation

konrad-krol-incubly:
## Purpose

The Automation API gained `area` on `DocumentationSpec` (gravitee-io/gravitee-api-management#19009) which is  `TOP_NAVBAR` by default.

`HOMEPAGE` marks the page as the portal's homepage.

This PR adds the matching `area` support to GKO's `Documentation` CRD and passes it through to the Automation API.

## Summary of changes

**The field**
A new `PageArea` enum with possible values: `TOP_NAVBAR`, `HOMEPAGE` and an optional `area` on `docs.Type`.

There is no `kubebuilder:default` on purpose. Setting one would add `area` to every existing `Documentation` spec, change its hash, and trigger a pointless reconcile of every page in the cluster. When `area` is unset GKO leaves it out of the request and lets APIM pick the default.

**Tests**
The unit tests call the real `CreateOrUpdate` against an `httptest` stub and check the request body, so `omitempty` is covered too: the value is sent for both areas, and the key is missing when unset. One test checks that changing the area changes the spec hash.

The integration tests cover the round trip, plus the backward compatible case: a page with no area reads back as `TOP_NAVBAR`, then gets promoted. A `withoutContext` test checks that the API server rejects an unknown area, since the enum lives on the CRD.

**Integration tests will pass when APIM changes are merged and deployed: gravitee-io/gravitee-api-management#19009**

### DIFF
--- a/test/unit/apim/documentation_test.go
+++ b/test/unit/apim/documentation_test.go
@@ -1,0 +1,64 @@
+// Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//         http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apim_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	documentation "github.com/gravitee-io/gravitee-kubernetes-operator/api/model/docs"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/utils"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
+)
+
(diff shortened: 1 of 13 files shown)


a-cordier (approved):

test/unit/apim/documentation_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func newPortal() *v1alpha1.Portal {
+	return &v1alpha1.Portal{
+		ObjectMeta: metav1.ObjectMeta{Name: "default-portal", Namespace: "default"},
+	}
+}
+
+var _ = Describe("Documentations service", func() {
```
benoitgravitee: This is nice but we don't to test GKO client. Although it sounds nicer to have this, it remains that integration test is enough to validate the field is populated, the client is indirectly tested, not ideal but sufficient

test/unit/apim/documentation_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	It("should omit the area when it is not set", func() {
+		body := captureDocumentationPut(newDocumentation(nil))
+
+		Expect(body).ToNot(HaveKey("area"))
+		Expect(body).To(HaveKeyWithValue("name", "Getting Started"))
+	})
+})
+
+var _ = Describe("Documentation spec", func() {
```
benoitgravitee, line 43: Nice

benoitgravitee (changes requested):

test/unit/apim/documentation_test.go
benoitgravitee: Maybe add an integration test for TOP_NAVBAR

konrad-krol-incubly: Done

benoitgravitee (approved):

PR #1778, opened 2026-08-12
feat: accept navigation structure on Portal with legacy fallback

konrad-krol-incubly:
## Purpose

The Automation API reworks `PortalSpec` navigation from a flat `navigation` array into a `structure` map keyed by portal area (gravitee-io/gravitee-api-management#19100). Legacy `navigation` is still accepted.

This PR adds `structure` to the `Portal` CRD and deprecates `navigation`.

## Summary of changes

**The field** - `structure.topNavbar`, an object rather than a map: a CRD map means `additionalProperties`, which cannot validate keys against an enum, so `HOMEPAGE: []` would apply cleanly and then fail server-side. Only `topNavbar` exists because it is the only area APIM accepts navigation for right now.

**Backward compatibility** - `navigation` is unchanged, so portals already in etcd sync as they do today. When both are set, `structure` wins and `navigation` is ignored, the same order APIM resolves them in.

**Deprecation** - `navigation` is marked deprecated in the CRD, so `kubectl explain` shows it, and the webhook warns on every apply that still uses it.

**Tests** - unit (precedence, spec hash), admission (warnings), integration (`structure` round trip). The existing `navigation` specs stay, so both paths are exercised.

**Integration tests will pass when APIM changes are merged and deployed: gravitee-io/gravitee-api-management#19100**

## Side effect:
 Fixes: https://gravitee.atlassian.net/browse/GKO-3021


konrad-krol-incubly:
- [x] wait with merge before https://github.com/gravitee-io/gravitee-api-management/pull/19100 lands in master and IT passes

crds/gravitee.io/gravitee.io_portals.yaml
```diff
@@ -95,6 +94,37 @@ spec:
...
+                        Top navbar entries as an ordered, flat list of paths.
+                        The order of entries in the list is preserved.
+                      items:
+                        description: NavigationPath is a single entry in a portal or API definition
+                          navigation tree.
+                        properties:
+                          displayName:
+                            description: Optional human-friendly label for this node.
+                            type: string
+                          order:
```
benoitgravitee: This field is not used at all, it not even part of the OAS spec. 
The index of the item in the list is used to populate the values in the DB. It is safe to remove it and would automatically solve: https://gravitee.atlassian.net/browse/GKO-3021 ;-)

test/integration/admission/portal/create_withContext_test.go
```diff
@@ -58,4 +58,39 @@ var _ = Describe("Validate create", labels.WithContext, func() {
...
+		prtl := fixture.
+			Builder().
+			WithPortal(constants.PortalStructureFile).
+			Build()
+
+		prtl.Portal.Spec.Navigation = prtl.Portal.Spec.Structure.TopNavbar
+
+		warnings, _ := admissionCtrl.ValidateCreate(ctx, prtl.Portal)
+
+		Expect(warnings).To(ContainElement(ContainSubstring("ignored because spec.structure is set")))
```
benoitgravitee: We have other cases like this, we treat them differently.
We expect a failure when using both deprecated and non-deprecated value.
The reason behind it is that we want TF and GKO to be consistent and in TF we don't have warnings. So we must fail hence the "exclusive or" kind of validation

internal/admission/portal/validate.go
```diff
@@ -46,6 +48,26 @@ func validateCreate(ctx context.Context, obj runtime.Object) *errors.AdmissionEr
 	return errs
 }
 
+func warnOnDeprecatedNavigation(prtl *v1alpha1.Portal) *errors.AdmissionErrors {
```
benoitgravitee: see my comment on tests

test/integration/admission/portal/create_withContext_test.go
benoitgravitee: If you use the deprecated field, a warning is expected though

test/unit/apim/portal_test.go
```diff
@@ -0,0 +1,90 @@
...
+	})
+
+	It("should fall back to the deprecated navigation when the structure is unset", func() {
+		spec := newPortalSpec(nil, navPath("/legacy"))
+
+		Expect(spec.Spec.GetTopNavbar()).To(ConsistOf(navPath("/legacy")))
+	})
+
+	// APIM ignores navigation as soon as structure is set, so GKO resolves the same way.
+	It("should ignore the deprecated navigation when the structure is set", func() {
```
benoitgravitee: see "XOR" comment above.

benoitgravitee (changes requested):

crds/gravitee.io/gravitee.io_portals.yaml
konrad-krol-incubly: Removed, also from portalNavigation.

test/integration/admission/portal/create_withContext_test.go
konrad-krol-incubly: Done - severe error when both fields are present, warning kept for legacy field only.

konrad-krol-incubly: BTW in this case APIM automation-api should behave the same way -> when both fields are present -> it should reject request.

@benoitgravitee do you think this is correct assumption?

In this case I need to adjust APIM implementation.

internal/admission/portal/validate.go
konrad-krol-incubly: Done - severe error now.

test/unit/apim/portal_test.go
konrad-krol-incubly: updated

crds/gravitee.io/gravitee.io_portals.yaml
benoitgravitee: No that would make a BC.
Keep it in portalNavigation, it is fine, the whole stuff is deprecated.

test/integration/admission/portal/create_withContext_test.go
benoitgravitee: Yes. It must be done in the API as well.

konrad-krol-incubly: ok, I will adjust APIM

crds/gravitee.io/gravitee.io_portals.yaml
konrad-krol-incubly: ok, you are right -> i left that to not introduce breaking changes

benoitgravitee (approved):

PR #1781, opened 2026-08-12
fix(subscription): do not panic when validating a subscription to an API without context

paulatulis:
https://gravitee.atlassian.net/browse/APIM-14905

### Problem
`validateContextRefs` called `GetName()` on the API's and application's `contextRef` without checking whether either exists. `ContextRef()` returns `spec.contextRef` verbatim, so an unset context yields a nil pointer wrapped in a non-nil interface, and the admission webhook panics instead of returning a validation error:

    admission webhook "v1alpha1.gravitee.io.subscription" denied the request:
    panic: runtime error: invalid memory address or nil pointer dereference

The user gets a stack trace instead of a reason, and nothing is stored.

### Fix
Guard each read with the existing `HasContext()` on `ContextAwareObject`, and give each failure its own actionable message:

- API has no `contextRef` → "unable to subscribe to API [x] because it does not reference a management context"
- application has no `contextRef` → equivalent message for the application
- both present but different → the existing "management contexts must match" error, unchanged

The mismatch branch is byte-identical to master, so the existing `create_withContext_and_contextMissmatch_test.go` is unaffected.

### Behaviour note for the reviewer
The ticket asks for "if neither side has a context, let the subscription through". This PR **rejects** that case instead, for two reasons:

- `Application.contextRef` is required by its CRD and `ApiDefinition`/`ApiV4Definition` `contextRef` is optional, so "neither" is unreachable in practice — the case that actually panics is API-without-context plus application-with-one, matching the reported stack trace.
- The subscription controller builds its APIM client from the *API's* contextRef (`controllers/apim/subscription/internal/update.go:50`), so admitting a subscription to a context-less API would create a resource that can never reconcile.

Happy to flip to "skip the check" if you'd rather match the ticket's literal wording.

### Tests
- `test/integration/admission/subscription/create_withContext_and_apiWithoutContext_test.go`
  — API with no context, application with one, asserting the admission error. The subscription is handed to the admission controller directly and never applied: the reconciler cannot build an APIM client for a context-less API, so the resource would never reach a terminal state.
- `internal/admission/subscription/validate_test.go` — table-driven, all six context combinations. Note it lives in package because `validateContextRefs` is unexported, and CI's `make unit` only runs `test/unit/...`, so it does **not** execute in CI as things stand. `go test ./internal/...` is green today if we want to extend that target (separately — it also revives two dormant `internal/k8s` tests).

### DIFF
--- a/internal/admission/subscription/validate.go
+++ b/internal/admission/subscription/validate.go
@@ -146,7 +146,7 @@ func validateCreate(ctx context.Context, obj runtime.Object) *errors.AdmissionEr
 		return errs
 	}
 
-	errs.Add(validateContextRefs(api, app))
+	errs.Add(ValidateContextRefs(api, app))
 	if errs.IsSevere() {
 		return errs
 	}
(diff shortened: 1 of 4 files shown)


benoitgravitee (approved):

internal/admission/subscription/validate_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/model/refs"
+	"github.com/gravitee-io/gravitee-kubernetes-operator/api/v1alpha1"
+)
+
+func contextRef(namespace, name string) *refs.NamespacedName {
+	return &refs.NamespacedName{Namespace: namespace, Name: name}
+}
+
+func TestValidateContextRefs(t *testing.T) {
```
benoitgravitee: this test should be test/unit and be written in ginkgo/gomega style to match other tests

benoitgravitee (changes requested):

internal/admission/subscription/validate_test.go
paulatulis: Addressed

benoitgravitee (approved):

PR #1794, opened 2026-08-16
fix app missing cotext panic error

nickdallavalentina:
ref Issue: https://github.com/gravitee-io/issues/issues/11722

The function called api.ContextRef() / app.ContextRef() unconditionally. When an API or Application had no context set, ContextRef() returned nil, and the subsequent comparison/formatting caused a panic during admission validation.

Fix: check HasContext() first a mismatch where one has a context and the other doesn't is now caught without dereferencing; the two ContextRef() calls only happen when both sides actually have a context. Also added a contextRefString() helper so the error message prints "none" instead of risking a nil format.


nickdallavalentina:
the fix is no longer needed. The same issue (panic when validating a subscription context ref that's nil) was already addressed and merged via #1781

PR #1797, opened 2026-08-21
fix admission webhook endpoint set to inherit

nickdallavalentina:
bugfix for [11747](https://github.com/gravitee-io/issues/issues/11747) 

admission webhook for Definition does not check that an endpoint set to inherit configuration

this function check that any endpoint with inheritConfiguration = true actually has a shared configuration in its EndpointGroup. If not, it adds a severe validation error in the admission webhook

### DIFF
--- a/internal/admission/api/v4/flows.go
+++ b/internal/admission/api/v4/flows.go
@@ -76,6 +76,38 @@ func validateApiEndpointGroups(api *apiV4.Api) *errors.AdmissionErrors {
 		if eg.HttpClientOptions != nil {
 			errs.MergeWith(validateHttpClientOptions(eg.HttpClientOptions, eg.Name, api.Name))
 		}
+		errs.MergeWith(validateEndpointsInheritConfiguration(eg, api.Name))
+	}
+
+	return errs
+}
+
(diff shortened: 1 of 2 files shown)


benoitgravitee (changes requested):
Thank you for your contribution this is well appreciated.
I will submit this change internally and will get back to you.

Meanwhile this needs to be tested. Add a test in tests/integration/admission/api/v2 and tests/integration/admission/api/v4 to validate your fix. You will find very similar test case that will help you create your own.

Cheers

nickdallavalentina:
> Thank you for your contribution this is well appreciated. I will submit this change internally and will get back to you.
> 
> Meanwhile this needs to be tested. Add a test in tests/integration/admission/api/v2 and tests/integration/admission/api/v4 to validate your fix. You will find very similar test case that will help you create your own.
> 
> Cheers

Thanks I added a test for v4, where this validation actually lives. I quick search at the v2 as well  its EndpointGroup has no sharedConfiguration field and also I checked the v2 webhook doesn't do local EndpointGroup validation at all. So this fix doesn't carry over to v2 as-is

benoitgravitee (approved):

PR #1804, opened 2026-08-24
feat: add PortalTheme support for the next-gen portal

konrad-krol-incubly:
## Purpose

The Automation API gained a portal theme resource, applied per environment, plus `activeThemeHrid` on `PortalSpec` so a portal can declare the theme it uses (gravitee-io/gravitee-api-management#19323).

This PR adds the `PortalTheme` CRD and `Portal.spec.themeRef`.

## Summary of changes

**The CRD** - mirrors `PortalThemeSpec` field for field:

- `name` - display name of the theme
- `definition` (required) - `customCss`, `font.fontFamily`, and `color`: `primary`, `secondary`, `tertiary`, `error`, `background.page` and `background.card`
- `logo`, `optionalLogo`, `favicon`, `backgroundImage` - images, inline data URIs
- `contextRef` - themes are environment-scoped, like `Portal`

**The link** - `themeRef` points at a `PortalTheme` and is sent to APIM as the theme HRID. Setting it activates the theme and disables the others; removing it deactivates it and clears `activeThemeHrid`.

**Delete guard** - APIM refuses to delete a theme a portal still has active, so the operator blocks the deletion first and names the portals to change. Deletion completes on its own once nothing uses the theme.

**Tests** - unit: spec hash over `definition` and `themeRef`. Admission: `contextRef` missing or unresolvable, dry-run rejection. Integration: theme round trip, `themeRef` reaching `activeThemeHrid`, delete guard releasing once unused.

Integration tests will pass when APIM changes are merged and deployed: gravitee-io/gravitee-api-management#19323 -> already passes

## Note for the reviewer

No `enabled` field mapped - it becomes legacy on APIM side (cannot address portal to theme one to many association).

## Blockers

- [x] Wait for PR with drift detection mechanism to be merged first!


test/internal/integration/fixture/decode.go
```diff
@@ -47,6 +47,7 @@ var (
 	portalListingKind      = v1alpha1.GroupVersion.WithKind("PortalListing")
 	portalLinkKind         = v1alpha1.GroupVersion.WithKind("PortalLink")
 	documentationKind      = v1alpha1.GroupVersion.WithKind("Documentation")
+	portalThemeKind        = v1alpha1.GroupVersion.WithKind("Theme")
```
MateuszJozwiak-Incubly: Shouldn't it be "PortalTheme" here?

konrad-krol-incubly: You are right, I didn't catched that because it was explicitly provided in my local CRDs I used for testing. Fixed

konrad-krol-incubly:
Tested drift detection mechanism and it works fine.

When I edit some properties from webconsole to cause drift and then apply from GKO side - its rejected:

```
Error from server (Forbidden): admission webhook "v1alpha1.gravitee.io.portaltheme" denied the request: 
drift detected:
definition:
  color:
    primary: "#6C2BD9" != "#FF0000"
```

internal/admission/portaltheme/validate.go
```diff
@@ -0,0 +1,69 @@
...
+		return errs
+	}
+
+	status, err := apimClient.PortalThemes.DryRunCreateOrUpdate(cp)
+	if err != nil {
+		errs.AddSevere(err.Error())
+		return errs
+	}
+
+	for _, severe := range status.Errors.Severe {
```
benoitgravitee: You could use errs.MergeWith here no ?

api/v1alpha1/portal_types.go
```diff
@@ -134,6 +139,31 @@ func (p *Portal) ContextRef() core.ObjectRef {
...
+	if p.Spec.Theme == nil {
+		return ""
+	}
+
+	ns := p.Spec.Theme.Namespace
+	if ns == "" {
+		ns = p.GetNamespace()
+	}
+
+	ref := refs.NewNamespacedName(ns, p.Spec.Theme.Name)
```
benoitgravitee: Use NewNamespaceNameFromObject here

internal/apim/service/portals.go
```diff
@@ -48,7 +48,11 @@ func (svc *Portals) createOrUpdate(
 	url := svc.AutomationTarget("portals").
 		WithQueryParam("dryRun", strconv.FormatBool(dryRun))
 
-	dto := model.ToPortalDTO(prtl.Spec.Type, refs.NewNamespacedNameFromObject(prtl).HRID())
+	dto := model.ToPortalDTO(
+		prtl.Spec.Type,
+		refs.NewNamespacedNameFromObject(prtl).HRID(),
+		prtl.ActiveThemeHRID(),
```
benoitgravitee: controllers/apim/portal/internal/update.go should do resolve the theme, and only then pass it to internam/apim/service. Here you don't know if the theme exist in GKO, webhook remains an option. If not portal should have a resolveRef condition set to false

test/integration/admission/portal/create_withContext_test.go
```diff
@@ -99,4 +99,31 @@ var _ = Describe("Validate create", labels.WithContext, func() {
...
+			Builder().
+			WithPortal(constants.PortalWithThemeFile).
+			Build()
+
+		_, err := admissionCtrl.ValidateCreate(ctx, prtl.Portal)
+
+		if err != nil {
+			Expect(err.Error()).ToNot(ContainSubstring("PortalTheme is supported"))
+		}
+	})
```
benoitgravitee, line 148: Add should fail when themeRef is not resolved

internal/admission/portaltheme/ctrl.go
```diff
@@ -0,0 +1,62 @@
...
+func (a AdmissionCtrl) ValidateCreate(
+	ctx context.Context, obj *v1alpha1.PortalTheme,
+) (admission.Warnings, error) {
+	return validateCreate(ctx, obj).Map()
+}
+
+func (a AdmissionCtrl) ValidateDelete(
+	_ context.Context, _ *v1alpha1.PortalTheme,
+) (admission.Warnings, error) {
+	return admission.Warnings{}, nil
```
benoitgravitee: You need to check that portal does not use a theme before it is removed

test/integration/admission/portaltheme/create_withContext_test.go
```diff
@@ -0,0 +1,78 @@
...
+		thm := fixture.
+			Builder().
+			WithPortalTheme(constants.PortalThemeFile).
+			WithContext(constants.ContextWithSecretFile).
+			Build()
+
+		thm.PortalTheme.Spec.Name = ""
+		thm.PortalTheme.Spec.Definition = portaltheme.Definition{}
+
+		Eventually(func() error {
```
benoitgravitee, line 73: You need to add update and delete test for the sake of it.

test/integration/portal/create_withContext_test.go
```diff
@@ -129,4 +129,32 @@ var _ = Describe("Create", labels.WithContext, func() {
 			return assert.ContainsInOrder("Portal navigation", expectedPaths, paths)
 		}, timeout, interval).Should(Succeed(), fixtures.Portal.Name)
 	})
+
+	It("should create portal in APIM with its active theme", func() {
```
benoitgravitee, line 134: Need a fail test here. Webhooks are not called in those test

benoitgravitee (changes requested):
portal -> theme dependency is not managed correctly, need to resolve in the controller as well before sending to the service layer. Hence some tests are missing.

internal/admission/portaltheme/validate.go
konrad-krol-incubly: Done.

api/v1alpha1/portal_types.go
konrad-krol-incubly: - Done - used the ref-taking variant, since we only have the ref here, not the object.

internal/apim/service/portals.go
konrad-krol-incubly: - Done - the controller resolves first and passes the resolved theme down; failure sets ResolvedRefs to false and requeues.
- Also added a watch so the portal recovers as soon as the theme lands.

test/integration/admission/portal/create_withContext_test.go
konrad-krol-incubly: - Added.
- Also fixed the neighbouring accept test, which never actually asserted acceptance.

internal/admission/portaltheme/ctrl.go
konrad-krol-incubly: - Done - delete is now guarded.
- The webhook was already registered for DELETE, so it was being called and silently passing.

test/integration/admission/portaltheme/create_withContext_test.go
konrad-krol-incubly: Added - 4 update specs, 2 delete specs.

test/integration/portal/create_withContext_test.go
konrad-krol-incubly: Added - theme missing means unresolved and nothing synced, then creating the theme syncs the portal.

benoitgravitee (approved):

PR #1816, opened 2026-09-04
feat: support visibility on next-gen portal navigation entries

konrad-krol-incubly:
## Purpose

The Automation API now accepts `visibility` (`PUBLIC` / `PRIVATE`) on portal navigation entries (gravitee-io/gravitee-api-management#19555). Before this everything it created was always `PUBLIC`, so you could set visibility in the Web Console but not from a CRD.

This PR adds `visibility` to the portal CRDs.

## Summary of changes

**The field**

A new `Visibility` enum with `PUBLIC` and `PRIVATE`, and an optional `visibility` on `Portal.spec.structure.topNavbar[]`, the deprecated `Portal.spec.navigation[]`, `ApiV4Definition.spec.portalNavigation[]`, `PortalListing.spec.apis[]`, `Documentation.spec` and `PortalLink.spec`.

There is no `kubebuilder:default` on purpose. APIM works out the value itself when we don't send one: it takes the visibility of the parent folder, and only falls back to `PUBLIC` at the root. If GKO always sent `PUBLIC`, anything filed under a private folder would be rejected unless the user repeated `visibility: PRIVATE` on every child, and entries that inherit `PRIVATE` today would turn public after an upgrade. So when `visibility` is unset GKO leaves it out of the request.

**Drift**

A new `ignore-unset` tag. If the CRD doesn't set `visibility` there is nothing to compare, so it never reports drift. If it does set it, it is compared as usual. `ignore-remote:PUBLIC,PRIVATE` would have ignored everything, including a real change made in the console.

**Tests**

Unit tests for the new tag, plus an unset-visibility case in each drift suite. Integration tests check that a value we send comes back, that an unset one stays unset, and that a page under a private folder with no `visibility` of its own ends up `PRIVATE`.

**Integration tests will pass when APIM changes are merged and deployed: gravitee-io/gravitee-api-management#19555**

## Blockers

- [x] gravitee-io/gravitee-api-management#19555


internal/apim/model/api_v4_dto.go
```diff
@@ -106,9 +107,10 @@ type APIV4ConsoleNotificationDTO struct {
 }
 
 type APIV4NavigationPathDTO struct {
-	Path        string  `json:"path"`
-	DisplayName *string `json:"displayName,omitempty"`
-	Order       *int32  `json:"order,omitempty"`
+	Path        string         `json:"path"`
+	DisplayName *string        `json:"displayName,omitempty"`
+	Order       *int32         `json:"order,omitempty"`
+	Visibility  nav.Visibility `json:"visibility,omitempty" drift:"ignore-unset"`
```
benoitgravitee: `ignore-remote:PUBLIC` would have been my choice.
It will consider fields to be equivalent if the remote contains this value.
The semantics is different.
We are trying to do the same here: ignore remote default.
Maybe you could:
* remove the `ignore-unset`
* rework `ignore-remote` to `ignore-remote-default` and change the implementation it Equivalent only when CRD is unset AND remote equals the value passed as the arg.
Your implementation and mine are too permissive, you allow from the remote, I allow all from the CRD.
Either way we can't the two.

benoitgravitee (changes requested):

internal/apim/model/api_v4_dto.go
MateuszJozwiak-Incubly: Adopted ignore-remote-default with the AND semantics you described — ignore-remote and ignore-unset are both gone, flowMode and documentation area moved over.           
                                                                                                                                                                           
  One catch on ignore-remote-default:PUBLIC for this field specifically: APIM resolves an omitted visibility as caller → parent → PUBLIC, not caller → PUBLIC. So an entry 
  left unset under a PRIVATE folder is stored as PRIVATE, and the tag would report that as permanent, unclearable drift.                                                   
                                                                                                                                                                           
  So instead of tagging it, the operator now resolves what APIM will store (own value → closest declared ancestor → PUBLIC) from the CRD's own path list, applied only on  
  the drift path — the payload still omits an undeclared visibility. portalNavigation and structure.topNavbar compare exactly now, no tag needed.

internal/apim/service/portallistings.go
```diff
@@ -93,6 +94,15 @@ func (svc *Listings) GetByHRID(portalHrid, listingHrid string) (*model.PortalLis
 	return listing, nil
 }
 
+// visibilityOf flattens the optional CRD visibility. Left unset, the field is
+// omitted from the payload so that APIM resolves it from the parent folder.
+func visibilityOf(v *nav.Visibility) nav.Visibility {
```
benoitgravitee: This should go away, caller should use `utils.SafeDereference()`

internal/drift/equivalences.go
```diff
@@ -50,18 +51,15 @@ func InitRegistry() {
...
 }
 
-// IgnoreRemoteArgs ignores the remote difference if the remote string is in the context.FuncArgs.
-func IgnoreRemoteArgs(crd any, remote any, context DriftContext) Equivalence {
+// IgnoreRemoteDefault ignores a difference only when the CRD leaves the field unset
+// AND the remote carries one of the server defaults listed in context.FuncArgs. It is
+// meant for fields APIM resolves itself when the payload omits them: a CRD that states
+// nothing cannot drift towards a value APIM was free to pick, while a CRD value that is
+// set is always compared, and a remote value the tag does not list is always drift.
+func IgnoreRemoteDefault(crd any, remote any, context DriftContext) Equivalence {
```
benoitgravitee, line 61: I think it should all values if no args is provided. That will simplify your tags and still make sense. Dev may choose to use restrictve values or not.

benoitgravitee (changes requested):

internal/apim/service/portallistings.go
MateuszJozwiak-Incubly: Done that :)

internal/drift/equivalences.go
MateuszJozwiak-Incubly: Agreed, that would be logical and more clean.

benoitgravitee (approved):

PR #1822, opened 2026-09-23
feat: reconcile catalog mcp server

a-cordier:
Adds `CatalogMcpServer`, an upstream MCP server registered in the AI Catalog (AIAM-232):

```yaml
kind: CatalogMcpServer
metadata:
  name: github
spec:
  contextRef:
    name: dev-ctx
  entityId: mcp-server.github   # immutable, the name authorization policies use
  connection:
    endpoint: https://api.githubcopilot.com/mcp/
    transport: HTTP
    auth:
      type: HEADER
      header:
        name: Authorization
        value: "[[ secret `github-mcp/token` ]]"
```

- Synced through `{automation}/aim/catalog/mcp-servers`, addressed by HRID.
- Admission does a dry run against the platform, so an unreachable server or an `entityId` that is already taken fails at `kubectl apply`.
- Drift detection covers updates. Credential values are never compared or echoed back.
- `status` reports what discovery found: server info, protocol version, tools, prompts and resources.

Examples:
  - examples/apim/catalog_mcp_server
  
🥼 Tested manually
---

- create catalog MCP server ✅ 
- create studio in the UI using this catalog server ✅ 
- call the studio proxy ✅ 
- delete the MCP server ✅ 
- call the studio proxy ✅ (that's expected because the API definition is self contained)


internal/admission/catalogmcpserver/drift.go
```diff
@@ -0,0 +1,58 @@
...
+
+func validateUpdate(
+	ctx context.Context,
+	oldObj *v1alpha1.CatalogMcpServer,
+	newObj *v1alpha1.CatalogMcpServer,
+) *errors.AdmissionErrors {
+	errs := validateCreate(ctx, newObj)
+	if errs.IsSevere() {
+		return errs
+	}
```
a-cordier: **improvement** — Every update, metadata-only ones included, re-runs the dry run, and the dry run discovers the upstream MCP server.

That covers the controller's own finalizer/`last-spec-hash` write, so each create reaches the upstream twice. While the upstream is down, `kubectl annotate ... gravitee.io/drift-detection=false` or a label change is rejected, and the controller's first metadata write fails the reconcile.

Application already guards this (`internal/admission/application/validate.go:88`). Suggest splitting `validateCreate` into the context check and the dry run, and running the dry run only when `oldObj.Spec.Hash() != newObj.Spec.Hash()`; drift keeps running on every update as for Application. Templated Secret rotation does not change the spec hash, but it never goes through admission anyway.

internal/admission/catalogmcpserver/validate.go
```diff
@@ -0,0 +1,89 @@
...
+				len(state.Errors.Severe) > 0 {
+				return errors.NewAdmissionErrorsFromStatus(state.Errors)
+			}
+		}
+	}
+
+	errs := errors.NewAdmissionErrors()
+	errs.AddSevere(err.Error())
+	return errs
+}
```
a-cordier: **improvement** — `refusalToAdmissionErrors` has no test, yet it decides what the author reads on the PR's headline path: an unreachable upstream or a taken `entityId` failing at `kubectl apply`.

It is pure (400 → decode the state body → one admission error per `errors.severe` entry, otherwise the raw error) but unexported, so no unit suite can reach it. Suggest moving the decoding to an exported helper in `internal/apim/model` (no import cycle: `internal/errors` only imports `api/model/status`) and table-testing it in `test/unit/apim`: a 400 with findings, a 400 whose body has none or does not parse, and a non-400 error. The reconcile path can then reuse it (see `update.go`).

controllers/apim/catalogmcpserver/internal/update.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	state, err := apimClient.CatalogMcpServers.CreateOrUpdate(srv)
+	if err != nil {
+		// The admission dry-run already refused bad payloads, so a 400 here is almost always
+		// the upstream being unreachable at reconcile time: keep it recoverable (backoff)
+		// rather than terminal until the next spec change.
+		if gerrors.IsBadRequest(err) {
+			return fmt.Errorf("catalog mcp server [%s] was refused by the automation api: %w", srv.GetName(), err)
+		}
```
a-cordier: **nit** — The comment's premise does not always hold, and a 400's findings never reach `status.errors`.

Admission does not filter bad payloads when the operator runs with `ENABLE_WEBHOOK=false`, nor when another server takes the `entityId` between admission and reconcile (the platform checks uniqueness at apply time). In those cases the reconcile retries under backoff indefinitely, and the platform's findings only show inside the raw condition message. Keeping it recoverable is fine; suggest rewording the comment and decoding the 400 body with the same helper admission uses, so `status.errors` carries the findings.

a-cordier (commented):
No blockers. The wire payload matches the aim module's automation contract (`AimCatalogMcpServerSpec`: required `hrid`, `type` discriminator, credentials redacted in every answer, `400` carrying the state with `errors.severe`), so the `drift:"ignore"` tags on `value` and `clientSecret` are right. Build, vet, the `apim`, `drift` and `predicate` unit suites and `make generate manifests reference` are clean on the head commit.

Three inline comments: the dry run reaching the upstream on metadata-only updates, the refusal decoding without a test, and the reconcile 400 path.

Residual risks: the discovered status is a snapshot from the last PUT (the platform no-ops an unchanged re-apply), and dry-run discovery plus the drift GET share the default 5 s client timeout inside a 10 s webhook timeout. No e2e coverage yet (GKO-3171, GKO-3173).

Security-sensitive: upstream credentials travel through templating to the platform; I checked they are not echoed by drift output, condition messages or 400 bodies.

This review was prepared with an AI reviewer; every comment was checked and approved by me before posting.

internal/apim/model/catalogmcpserver.go
```diff
@@ -0,0 +1,135 @@
...
+	if json.Unmarshal([]byte(serverError.Body), state) != nil || len(state.Errors.Severe) == 0 {
+		return status.Errors{}, false
+	}
+
+	return state.Errors, true
+}
+
+// ToCatalogMcpServerDTO maps the CRD onto the wire payload. It is the single mapping used by
+// the sync path and by drift detection.
+func ToCatalogMcpServerDTO(crd *v1alpha1.CatalogMcpServer) CatalogMcpServerDTO {
```
benoitgravitee: Many mapper use the trick of going through mapViaJSON, maybe you should consider it for the sake of being less verbose.

benoitgravitee (approved):

internal/apim/model/catalogmcpserver.go
a-cordier: got it, for that one, the thing is there is a little gotcha with the auth part, that's why we ended up there

a-cordier: I'll look if mapToJson makes thing better when I go with proxies, since that one has been tested in that exact state :)

